PR #7, opened 2024-03-26
Add methods to run Terraform Templates using TerraformRunner container APIs

putmanoj:
* Add `run` method to run terraform templates in async mode
* Add `run_async` method to run terraform templates
* `Response` class to represent TerraformRunner-Stack API response
* `ResponseAsync` class with `stop` method to cancel a TerraformRunner-Stack API Job. 
* Add spec the test terraform-runner methods

### DIFF
--- a/lib/terraform/runner.rb
+++ b/lib/terraform/runner.rb
@@ -7,16 +7,7 @@ require 'base64'
 module Terraform
   class Runner
     class << self
-      def server_url
-        ENV['TERRAFORM_RUNNER_URL'] || 'https://localhost:27000'
-      end
-
-      def server_token
-        # TODO: fix hardcoded token
-        ENV['TERRAFORM_RUNNER_TOKEN'] || 'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IlNodWJoYW5naSBTaW5naCIsImlhdCI6MTcwNjAwMDk0M30.46mL8RRxfHI4yveZ2wTsHyF7s2BAiU84aruHBoz2JRQ'
-      end
-
-      # Run a template, initiate stack creation (does wait to complete), via terraform-runner api
+      # Run a template, initiates terraform-runner job for running a template, via terraform-runner api
       #
       # @param input_vars [Hash] Hash with key/value pairs that will be passed as input variables to the
       #        terraform-runner run
@@ -28,7 +19,7 @@ module Terraform
       # @return [Terraform::Runner::ResponseAsync] Response object of terraform-runner create action
       def run_async(input_vars, template_path, tags: nil, credentials: [], env_vars: {})
         _log.info("In run_aysnc with #{template_path}")
-        response = run_create_stack(
+        response = create_stack_job(
           template_path,
           :input_vars  => input_vars,
           :tags        => tags,
@@ -38,7 +29,7 @@ module Terraform
         Terraform::Runner::ResponseAsync.new(response)
       end
 
-      # Runs a template, wait until it completes ,via terraform-runner api
+      # Runs a template, waits until it terraform-runner job completes, via terraform-runner api
       #
       # @param input_vars [Hash] Hash with key/value pairs that will be passed as input variables to the
       #        terraform-runner run
@@ -49,8 +40,8 @@ module Terraform
       #        terraform-runner run
       # @return [Terraform::Runner::Response] Response object with final result of terraform run
       def run(input_vars, template_path, tags: nil, credentials: [], env_vars: {})
-        _log.info("In run")
-        run_create_stack_and_wait_until_completes(
+        _log.info("Run template: #{template_path}")
+        create_stack_job_and_wait_until_completes(
           template_path,
           :input_vars  => input_vars,
           :tags        => tags,
@@ -59,15 +50,13 @@ module Terraform
         )
       end
 
-      def terraform_runner_client
-        # TODO: verify ssl
-        verify_ssl = false
-
-        RestClient::Resource.new(
-          server_url,
-          :headers    => {:authorization => "Bearer #{server_token}"},
-          :verify_ssl => verify_ssl
-        )
+      # Fetch terraform-runner job result/status by stack_id
+      #
+      # @param stack_id [String] stack_id from the terraforn-runner job
+      #
+      # @return [Terraform::Runner::Response] Response object with result of terraform run
+      def fetch_result_by_stack_id(stack_id)
+        retrieve_stack_job(stack_id)
       end
 
       def available?
@@ -96,7 +85,51 @@ module Terraform
         end
       end
 
-      def run_create_stack(
+
+      # =================================================
+      # TerraformRunner Stack-API interaction methods
+      # =================================================
+      private
+
+      def server_url
+        ENV['TERRAFORM_RUNNER_URL'] || 'https://localhost:27000'
+      end
+
+      def server_token
+        # TODO: fix hardcoded token
+        ENV['TERRAFORM_RUNNER_TOKEN'] || 'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IlNodWJoYW5naSBTaW5naCIsImlhdCI6MTcwNjAwMDk0M30.46mL8RRxfHI4yveZ2wTsHyF7s2BAiU84aruHBoz2JRQ'
+      end
+
+      def stack_job_interval_in_secs
+        ENV['TERRAFORM_RUNNER_STACK_JOB_CHECK_INTERVAL'].to_i
+      rescue
+        10
+      end
+
+      def stack_job_max_time_in_secs
+        ENV['TERRAFORM_RUNNER_STACK_JOB_MAX_TIME'].to_i
+      rescue
+        120
+      end
+
+      # create http client for terraform-runner rest-api
+      def terraform_runner_client
+        # TODO: verify ssl
+        verify_ssl = false
+
+        RestClient::Resource.new(
+          server_url,
+          :headers    => {:authorization => "Bearer #{server_token}"},
+          :verify_ssl => verify_ssl
+        )
+      end
+
+      def stack_tenant_id
+        '00000000-0000-0000-0000-000000000000'.freeze
+      end
+
+      # Create TerraformRunner Stack Job
+      def create_stack_job(
         template_path,
         input_vars: [],
         tags: nil,
@@ -104,9 +137,8 @@ module Terraform
         env_vars: {},
         name: "stack-#{rand(36**8).to_s(36)}"
       )
-        _log.info("In run_create_stack")
-        # TODO: fix hardcoded tenant_id
-        tenant_id = 'c158d710-d91c-11ed-9fee-d93323035b4e'
+        _log.info("create_stack_job for #{template_path}")
+        tenant_id = stack_tenant_id
 
         Tempfile.create(%w[opentofu-runner-payload .zip]) do |zip_file|
           create_zip_file_from_directory(zip_file.path, template_path)
@@ -122,7 +154,7 @@ module Terraform
               :parameters      => convert_to_cam_parameters(input_vars)
             }
           )
-          _log.info("Payload:>\n, #{payload}")
+          # _log.debug("Payload:>\n, #{payload}")
           http_response = terraform_runner_client['api/stack/create'].post(
             payload, :content_type => 'application/json'
           )
@@ -132,7 +164,8 @@ module Terraform
         end
       end
 
-      def run_retrieve_stack_by_id(stack_id)
+      # Retrieve TerraformRunner Stack Job details
+      def retrieve_stack_job(stack_id)
         payload = JSON.generate(
           {
             :stack_id => stack_id
@@ -145,19 +178,20 @@ module Terraform
         Terraform::Runner::Response.parsed_response(http_response)
       end
 
+      # Wait for TerraformRunner Stack Job to complete
       def wait_until_completes(stack_id)
-        interval_in_secs = 10
-        max_time_in_secs = 60
+        interval_in_secs = stack_job_interval_in_secs
+        max_time_in_secs = stack_job_max_time_in_secs
 
         response = nil
         Timeout.timeout(max_time_in_secs) do
-          _log.info("Starting wait for stack/#{stack_id} completes ...")
+          _log.info("Starting wait for terraform-runner/stack/#{stack_id} completes ...")
           i = 0
           loop do
             _log.info(i)
             i += 1
 
-            response = run_retrieve_stack_by_id(stack_id)
+            response = retrieve_stack_job(stack_id)
 
             _log.info("status: #{response.status}")
 
@@ -177,6 +211,8 @@ module Terraform
               break
             end
             _log.info("============\n #{response.message} \n============")
+
+            # sleep interval
             _log.info("Sleep for #{interval_in_secs} secs")
             sleep interval_in_secs
 
@@ -187,7 +223,8 @@ module Terraform
         response
       end
 
-      def run_create_stack_and_wait_until_completes(
+      # Create TerraformRunner Stack Job, wait until completes
+      def create_stack_job_and_wait_until_completes(
         template_path,
         input_vars: [],
         tags: nil,
@@ -195,8 +232,8 @@ module Terraform
         env_vars: {},
         name: "stack-#{rand(36**8).to_s(36)}"
       )
-        _log.info("In run_create_stack_and_wait_until_completes")
-        response = run_create_stack(
+        _log.info("create_stack_job_and_wait_until_completes for #{template_path}")
+        response = create_stack_job(
           template_path,
           :input_vars  => input_vars,
           :tags        => tags,
@@ -207,6 +244,7 @@ module Terraform
         wait_until_completes(response.stack_id)
       end
 
+      # create zip from directory
       def create_zip_file_from_directory(zip_file_path, template_path)
         dir_path = template_path # directory to be zipped
         dir_path = path[0...-1] if dir_path.end_with?('/')
--- a/lib/terraform/runner.rb
+++ b/lib/terraform/runner.rb
@@ -83,12 +83,11 @@ module Terraform
       private
 
       def server_url
-        ENV['TERRAFORM_RUNNER_URL'] || 'https://localhost:27000'
+        ENV['TERRAFORM_RUNNER_URL'] || 'https://opentofu-runner:27000'
       end
 
       def server_token
-        # TODO: fix hardcoded token
-        ENV['TERRAFORM_RUNNER_TOKEN'] || 'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IlNodWJoYW5naSBTaW5naCIsImlhdCI6MTcwNjAwMDk0M30.46mL8RRxfHI4yveZ2wTsHyF7s2BAiU84aruHBoz2JRQ'
+        ENV.fetch('TERRAFORM_RUNNER_TOKEN', nil)
       end
 
       def stack_job_interval_in_secs
--- a/lib/terraform/runner.rb
+++ b/lib/terraform/runner.rb
@@ -83,7 +83,7 @@ module Terraform
       private
 
       def server_url
-        ENV['TERRAFORM_RUNNER_URL'] || 'https://opentofu-runner:27000'
+        ENV.fetch('TERRAFORM_RUNNER_URL', 'https://opentofu-runner:27000')
       end
 
       def server_token
@@ -91,15 +91,11 @@ module Terraform
       end
 
       def stack_job_interval_in_secs
-        ENV['TERRAFORM_RUNNER_STACK_JOB_CHECK_INTERVAL'].to_i
-      rescue
-        10
+        ENV.fetch('TERRAFORM_RUNNER_STACK_JOB_CHECK_INTERVAL', 10).to_i
       end
 
       def stack_job_max_time_in_secs
-        ENV['TERRAFORM_RUNNER_STACK_JOB_MAX_TIME'].to_i
-      rescue
-        120
+        ENV.fetch('TERRAFORM_RUNNER_STACK_JOB_MAX_TIME', 120).to_i
       end
 
       # Create to paramaters as used by terraform-runner api
--- a/lib/terraform/runner.rb
+++ b/lib/terraform/runner.rb
@@ -133,6 +133,10 @@ module Terraform
         '00000000-0000-0000-0000-000000000000'.freeze
       end
 
+      def json_post_arguments(payload)
+        return JSON.generate(payload), "Content-Type" => "application/json".freeze
+      end
+
       # Create TerraformRunner Stack Job
       def create_stack_job(
         template_path,
@@ -147,20 +151,18 @@ module Terraform
         encoded_zip_file = encoded_zip_from_directory(template_path)
 
         # TODO: use tags,env_vars
-        payload = JSON.generate(
-          {
-            :cloud_providers => credentials,
-            :name            => name,
-            :tenantId        => tenant_id,
-            :templateZipFile => encoded_zip_file,
-            :parameters      => convert_to_cam_parameters(input_vars)
-          }
-        )
+        payload = {
+          :cloud_providers => credentials,
+          :name            => name,
+          :tenantId        => tenant_id,
+          :templateZipFile => encoded_zip_file,
+          :parameters      => convert_to_cam_parameters(input_vars)
+        }
         # _log.debug("Payload:>\n, #{payload}")
+
         http_response = terraform_runner_client.post(
           "api/stack/create",
-          payload,
-          "Content-Type" => "application/json"
+          *json_post_arguments(payload)
         )
         _log.debug("==== http_response.body: \n #{http_response.body}")
         _log.info("stack_job for template: #{template_path} running ...")
@@ -169,11 +171,9 @@ module Terraform
 
       # Retrieve TerraformRunner Stack Job details
       def retrieve_stack_job(stack_id)
-        payload = JSON.generate({:stack_id => stack_id})
         http_response = terraform_runner_client.post(
           "api/stack/retrieve",
-          payload,
-          "Content-Type" => "application/json"
+          *json_post_arguments({:stack_id => stack_id})
         )
         _log.info("==== Retrieve Stack Response: \n #{http_response.body}")
         Terraform::Runner::Response.parsed_response(http_response)
@@ -181,11 +181,9 @@ module Terraform
 
       # Cancel/Stop running TerraformRunner Stack Job
       def cancel_stack_job(stack_id)
-        payload = JSON.generate({:stack_id => stack_id})
         http_response = terraform_runner_client.post(
           "api/stack/cancel",
-          payload,
-          "Content-Type" => "application/json"
+          *json_post_arguments({:stack_id => stack_id})
         )
         _log.info("==== Cancel Stack Response: \n #{http_response.body}")
         Terraform::Runner::Response.parsed_response(http_response)
--- a/lib/terraform/runner.rb
+++ b/lib/terraform/runner.rb
@@ -1,0 +1,227 @@
+require 'rest-client'
+require 'timeout'
+require 'tempfile'
+require 'zip'
+require 'base64'
+
+module Terraform
+  class Runner
+    class << self
+      def server_url
+        ENV['TERRAFORM_RUNNER_URL'] || 'https://localhost:27000'
+      end
+
+      def server_token
+        # TODO: fix hardcoded token
+        ENV['TERRAFORM_RUNNER_TOKEN'] || 'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IlNodWJoYW5naSBTaW5naCIsImlhdCI6MTcwNjAwMDk0M30.46mL8RRxfHI4yveZ2wTsHyF7s2BAiU84aruHBoz2JRQ'
+      end
+
+      # Run a template, initiate stack creation (does wait to complete), via terraform-runner api
+      #
+      # @param input_vars [Hash] Hash with key/value pairs that will be passed as input variables to the
+      #        terraform-runner run
+      # @param template_path [String] Path to the template we will want to run
+      # @param tags [Hash] Hash with key/values pairs that will be passed as tags to the terraform-runner run
+      # @param credentials [Array] List of Authentication object ids to provide to the terraform run
+      # @param env_vars [Hash] Hash with key/value pairs that will be passed as environment variables to the
+      #        terraform-runner run
+      # @return [Terraform::Runner::ResponseAsync] Response object of terraform-runner create action
+      def run_async(input_vars, template_path, tags: nil, credentials: [], env_vars: {})
+        _log.info("In run_aysnc with #{template_path}")
+        response = run_create_stack(
+          template_path,
+          :input_vars  => input_vars,
+          :tags        => tags,
+          :credentials => credentials,
+          :env_vars    => env_vars
+        )
+        Terraform::Runner::ResponseAsync.new(response)
+      end
+
+      # Runs a template, wait until it completes ,via terraform-runner api
+      #
+      # @param input_vars [Hash] Hash with key/value pairs that will be passed as input variables to the
+      #        terraform-runner run
+      # @param template_path [String] Path to the template we will want to run
+      # @param tags [Hash] Hash with key/values pairs that will be passed as tags to the terraform-runner run
+      # @param credentials [Array] List of Authentication object ids to provide to the terraform run
+      # @param env_vars [Hash] Hash with key/value pairs that will be passed as environment variables to the
+      #        terraform-runner run
+      # @return [Terraform::Runner::Response] Response object with final result of terraform run
+      def run(input_vars, template_path, tags: nil, credentials: [], env_vars: {})
+        _log.info("In run")
+        run_create_stack_and_wait_until_completes(
+          template_path,
+          :input_vars  => input_vars,
+          :tags        => tags,
+          :credentials => credentials,
+          :env_vars    => env_vars
+        )
+      end
+
+      def terraform_runner_client
+        # TODO: verify ssl
+        verify_ssl = false
+
+        RestClient::Resource.new(
+          server_url,
+          :headers    => {:authorization => "Bearer #{server_token}"},
+          :verify_ssl => verify_ssl
+        )
+      end
+
+      def available?
+        return @available if defined?(@available)
+
+        response = terraform_runner_client['api/terraformjobs/count'].get
+        @available = response.code == 200
+      rescue
+        @available = false
+      end
+
+      # Run a template, initiate stack creation (does wait to complete), via terraform-runner api
+      #
+      # @param vars [Hash] Hash with key/value pairs that will be passed as input variables to the
+      #        terraform-runner run
+      # @return [Array] Array of {:name,:value}
+      def convert_to_cam_parameters(vars)
+        parameters = []
+        vars.each do |key, value|
+          parameters.push(
+            {
+              :name  => key,
+              :value => value
+            }
+          )
+        end
+      end
+
+      def run_create_stack(
+        template_path,
+        input_vars: [],
+        tags: nil,
+        credentials: [],
+        env_vars: {},
+        name: "stack-#{rand(36**8).to_s(36)}"
+      )
+        _log.info("In run_create_stack")
+        # TODO: fix hardcoded tenant_id
+        tenant_id = 'c158d710-d91c-11ed-9fee-d93323035b4e'
+
+        Tempfile.create(%w[opentofu-runner-payload .zip]) do |zip_file|
+          create_zip_file_from_directory(zip_file.path, template_path)
+          encoded_zip_file = Base64.encode64(File.binread(zip_file.path))
+
+          # TODO: use tags,env_vars
+          payload = JSON.generate(
+            {
+              :cloud_providers => credentials,
+              :name            => name,
+              :tenantId        => tenant_id,
+              :templateZipFile => encoded_zip_file,
+              :parameters      => convert_to_cam_parameters(input_vars)
+            }
+          )
+          _log.info("Payload:>\n, #{payload}")
+          http_response = terraform_runner_client['api/stack/create'].post(
+            payload, :content_type => 'application/json'
+          )
+          _log.info("==== http_response.body: \n #{http_response.body}")
+          _log.info("stack_job for template: #{template_path} running ...")
+          Terraform::Runner::Response.parsed_response(http_response)
+        end
+      end
+
+      def run_retrieve_stack_by_id(stack_id)
+        payload = JSON.generate(
+          {
+            :stack_id => stack_id
+          }
+        )
+        http_response = terraform_runner_client['api/stack/retrieve'].post(
+          payload, :content_type => 'application/json'
+        )
+        _log.info("==== http_response.body: \n #{http_response.body}")
+        Terraform::Runner::Response.parsed_response(http_response)
+      end
+
+      def wait_until_completes(stack_id)
+        interval_in_secs = 10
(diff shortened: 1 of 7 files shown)


lib/terraform/runner.rb
```diff
@@ -0,0 +1,293 @@
...
+      # =================================================
+      private
+
+      def server_url
+        ENV['TERRAFORM_RUNNER_URL'] || 'https://localhost:27000'
+      end
+
+      def server_token
+        # TODO: fix hardcoded token
+        ENV['TERRAFORM_RUNNER_TOKEN'] || 'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IlNodWJoYW5naSBTaW5naCIsImlhdCI6MTcwNjAwMDk0M30.46mL8RRxfHI4yveZ2wTsHyF7s2BAiU84aruHBoz2JRQ'
```
jrafanie, line 91: Maybe drop this hardcoded value and set it in your local environment for now?  We can also stub/mock it in tests if needed.  We don't want to accidentally post tokens.

lib/terraform/runner.rb
```diff
@@ -0,0 +1,293 @@
...
+        end
+        parameters
+      end
+
+      # create http client for terraform-runner rest-api
+      def terraform_runner_client
+        # TODO: verify ssl
+        verify_ssl = false
+
+        RestClient::Resource.new(
```
jrafanie, line 129: @Fryguy @agrare what do you think of adding another user of rest client?  No releases or activity for 4+ years.  Would faraday be a better option?

lib/terraform/runner.rb
```diff
@@ -0,0 +1,293 @@
...
+      def wait_until_completes(stack_id)
+        interval_in_secs = stack_job_interval_in_secs
+        max_time_in_secs = stack_job_max_time_in_secs
+
+        response = nil
+        Timeout.timeout(max_time_in_secs) do
+          _log.info("Starting wait for terraform-runner/stack/#{stack_id} completes ...")
+          i = 0
+          loop do
+            _log.info(i)
```
jrafanie, line 219: Some of these log lines might need to become debug lines.  We'll need to see what a full runner log looks like.

This one in particular looks like strict debug and not intended to be merged like this.

lib/terraform/runner/response_async.rb
```diff
@@ -0,0 +1,65 @@
...
+        @response = Terraform::Runner.fetch_result_by_stack_id(@stack_id)
+
+        @response
+      end
+
+      # @return [Terraform::Runner::Response, NilClass] Response object with all details about the Terraform run, or nil
+      #         if the Terraform is still running
+      def response
+        # return if running?
+        return @response if @response
```
jrafanie, line 38: Will wait to see the changes we discussed here, primarily around moving some of this logic to `running?` or another method called from there so we can use `running?`

agrare:
@putmanoj please rebase on master rather than merge master into this branch.

lib/terraform/runner.rb
putmanoj: move hardcode value into test

lib/terraform/runner.rb
putmanoj: using _log.debug

lib/terraform/runner/response_async.rb
putmanoj: these changes done, as suggested

putmanoj:
> @putmanoj please rebase on master rather than merge master into this branch.

ok

putmanoj:
> ~/.gitconfig

ok

lib/terraform/runner.rb
```diff
@@ -0,0 +1,293 @@
...
+require 'timeout'
+require 'tempfile'
+require 'zip'
+require 'base64'
+
+module Terraform
+  class Runner
+    class << self
+      def available?
+        return @available if defined?(@available)
```
jrafanie, line 11: ruby has a thing called double bang which can eliminate the need to check the `defined?` here.  If `@available` is `nil` or `false`, `!!@available` will be `false`.  If it's any other object, it will be `true`.

It uses negation twice to turn anything into a `true/false`. value and is best in `?` methods like this.   See: https://rubyquicktips.com/post/583755021/convert-anything-to-boolean

lib/terraform/runner.rb
```diff
@@ -0,0 +1,293 @@
...
+      def stack_job_interval_in_secs
+        ENV['TERRAFORM_RUNNER_STACK_JOB_CHECK_INTERVAL'].to_i
+      rescue
+        10
+      end
+
+      def stack_job_max_time_in_secs
+        ENV['TERRAFORM_RUNNER_STACK_JOB_MAX_TIME'].to_i
+      rescue
+        120
```
jrafanie, line 102:
```suggestion
        ENV.fetch('TERRAFORM_RUNNER_STACK_JOB_MAX_TIME', 120)
```

lib/terraform/runner.rb
```diff
@@ -0,0 +1,293 @@
...
+      end
+
+      def server_token
+        ENV.fetch('TERRAFORM_RUNNER_TOKEN', nil)
+      end
+
+      def stack_job_interval_in_secs
+        ENV['TERRAFORM_RUNNER_STACK_JOB_CHECK_INTERVAL'].to_i
+      rescue
+        10
```
jrafanie, line 96:
```suggestion
        ENV.fetch('TERRAFORM_RUNNER_STACK_JOB_CHECK_INTERVAL', 10)
```

lib/terraform/runner.rb
```diff
@@ -0,0 +1,293 @@
...
+        retrieve_stack_job(stack_id)
+      end
+
+      # =================================================
+      # TerraformRunner Stack-API interaction methods
+      # =================================================
+      private
+
+      def server_url
+        ENV['TERRAFORM_RUNNER_URL'] || 'https://opentofu-runner:27000'
```
jrafanie, line 86:
```suggestion
        ENV.fetch('TERRAFORM_RUNNER_URL', 'https://opentofu-runner:27000')
```

lib/terraform/runner.rb
```diff
@@ -0,0 +1,293 @@
...
+        name: "stack-#{rand(36**8).to_s(36)}"
+      )
+        _log.info("start stack_job for template: #{template_path}")
+        tenant_id = stack_tenant_id
+
+        # Temp Zip File
+        # zip_file_path = Tempfile.new(%w/tmp .zip/).path
+        zip_file_path = Tempfile.new(%w[tmp .zip]).path
+        create_zip_file_from_directory(zip_file_path, template_path)
+        zip_file_hash = Base64.encode64(File.binread(zip_file_path))
```
jrafanie, line 156: Should we also be encrypting the file contents?  could there there be sensitive data in the template?

lib/terraform/runner.rb
```diff
@@ -0,0 +1,293 @@
...
+        # _log.debug("Payload:>\n, #{payload}")
+        http_response = terraform_runner_client['api/stack/create'].post(
+          payload, :content_type => 'application/json'
+        )
+        _log.debug("==== http_response.body: \n #{http_response.body}")
+        _log.info("stack_job for template: #{template_path} running ...")
+        Terraform::Runner::Response.parsed_response(http_response)
+      ensure
+        # cleanup temp zip file
+        FileUtils.rm_rf(zip_file_path) if zip_file_path
```
jrafanie, line 177:
```suggestion
        FileUtils.rm_rf(zip_file_path) if zip_file_path && File.exist?(zip_file_path)
```

lib/terraform/runner.rb
```diff
@@ -0,0 +1,293 @@
...
+      # Cancel/Stop running TerraformRunner Stack Job
+      def cancel_stack_job(stack_id)
+        payload = JSON.generate(
+          {
+            :stack_id => stack_id
+          }
+        )
+        http_response = terraform_runner_client['api/stack/cancel'].post(
+          payload, :content_type => 'application/json'
+        )
```
jrafanie, line 204: Is it worthwhile to extract a method to build the post arguments?

```ruby
def build_post_arguments(payload_options)
  payload = JSON.generate(payload_options)
  return [payload, :content_type => 'application/json']
end

...
def cancel_stack_job(stack_id)
  post_arguments = build_post_arguments({:stack_id => stack_id})
  http_response = terraform_runner_client['api/stack/cancel'].post(*post_arguments)

  _log.info("==== Cancel Stack Response: \n #{http_response.body}")
  Terraform::Runner::Response.parsed_response(http_response)
end
```

lib/terraform/runner/response.rb
```diff
@@ -0,0 +1,56 @@
...
+                  :details, :created_at, :stack_job_start_time, :stack_job_end_time
+
+      # @return [String] Extracted attributes from the JSON response body object
+      def self.parsed_response(http_response)
+        data = JSON.parse(http_response.body)
+        _log.debug("data : #{data}")
+        Terraform::Runner::Response.new(
+          :stack_id             => data['stack_id'],
+          :stack_name           => data['stack_name'],
+          :status               => data['status'],
```
jrafanie, line 18: since we're checking for upcased SUCCESS/FAILED, should we do:

```suggestion
          :status               => data.fetch('status', '').upcase,
```

lib/terraform/runner.rb
```diff
@@ -0,0 +1,293 @@
...
+              _log.info(response.message)
+              break
+            end
+            _log.info("============\n stack_job/:#{stack_id} status=#{response.status} \n============")
+
+            # sleep interval
+            _log.debug("Sleep for #{interval_in_secs} secs")
+            sleep interval_in_secs
+
+            break unless i < 20
```
jrafanie, line 247: My brain reads this easier as `break if i >= 20`

lib/terraform/runner.rb
agrare: :+1: for faraday

lib/terraform/runner.rb
jrafanie: why are we both checking the iterations of the the loop and the max time?  do we need this line?

lib/terraform/runner.rb
```diff
@@ -0,0 +1,293 @@
...
+
+      # create zip from directory
+      def create_zip_file_from_directory(zip_file_path, template_path)
+        dir_path = template_path # directory to be zipped
+        dir_path = path[0...-1] if dir_path.end_with?('/')
+
+        _log.debug("Create #{zip_file_path}")
+        Zip::File.open(zip_file_path, Zip::File::CREATE) do |zipfile|
+          Dir.chdir(dir_path)
+          Dir.glob("**/*").reject { |fn| File.directory?(fn) }.each do |file|
```
jrafanie, line 283: I'd prefer selecting files vs. rejecting directories.

```suggestion
          Dir.glob("**/*").select { |f| File.file?(f) }.each do |file|
```

lib/terraform/runner/response.rb
```diff
@@ -0,0 +1,56 @@
...
+  class Runner
+    class Response
+      include Vmdb::Logging
+
+      attr_reader :stack_id, :stack_name, :status, :action, :message, :error_message,
+                  :details, :created_at, :stack_job_start_time, :stack_job_end_time
+
+      # @return [String] Extracted attributes from the JSON response body object
+      def self.parsed_response(http_response)
+        data = JSON.parse(http_response.body)
```
jrafanie, line 13: Question:  can the response body be empty/bad in the event of a 500 error or similar?  I'm wondering if `.body` will always be able to be `JSON.parse` against.

lib/terraform/runner.rb
```diff
@@ -0,0 +1,293 @@
...
+        parameters = []
+        vars&.each do |key, value|
+          parameters.push(
+            {
+              :name  => key,
+              :value => value
+            }
+          )
+        end
+        parameters
```
agrare, line 120: This looks like you can just use `.map` here
```suggestion
        vars&.map do |key, value|
          {
            :name  => key,
            :value => value
          }
        end
```

Not sure how much you care about returning `[]` if `vars` is `nil`, you could `return [] if vars.nil?` or `vars.to_h.map`

lib/terraform/runner/response.rb
```diff
@@ -0,0 +1,56 @@
...
+          :stack_id             => data['stack_id'],
+          :stack_name           => data['stack_name'],
+          :status               => data['status'],
+          :action               => data['action'],
+          :message              => data['message'],
+          :error_message        => data['error_message'],
+          :details              => data['details'],
+          :created_at           => data['created_at'],
+          :stack_job_start_time => data['stack_job_start_time'],
+          :stack_job_end_time   => data['stack_job_end_time']
```
jrafanie, line 25: do these get automatically converted to Date/time objects or should we do this too?  Can these be nil?

lib/terraform/runner/response_async.rb
```diff
@@ -0,0 +1,90 @@
...
+        # IF NOT SUCCESS,FAILED,CANCELLED
+        if status
+          return (
+            status.start_with?("SUCCESS", "FAILED") ||
+            # @response.status == "ERROR" ||
+            status == "CANCELLED"
+          )
+        end
+
+        false
```
jrafanie, line 86: Do we need `start_with`?   Can we do this?

```suggestion
        case status
        when "SUCCESS", "FAILED", "CANCELLED"
          true
        else
          false
        end
```

spec/lib/terraform/runner/data/hello-world/main.tf
```diff
@@ -0,0 +1,18 @@
...
+
+resource "null_resource" "null_resource_simple" {
+    provisioner "local-exec" {
+        command = "echo Hello '${var.name}'"
+    }
+}
+
+output "greeting" {
+  value = "Hello ${var.name}!"
+}
```
jrafanie, line 18: git is complaining about a missing newline at the end of the file.

lib/terraform/runner.rb
```diff
@@ -0,0 +1,293 @@
...
+        credentials: [],
+        env_vars: {},
+        name: "stack-#{rand(36**8).to_s(36)}"
+      )
+        _log.info("start stack_job for template: #{template_path}")
+        tenant_id = stack_tenant_id
+
+        # Temp Zip File
+        # zip_file_path = Tempfile.new(%w/tmp .zip/).path
+        zip_file_path = Tempfile.new(%w[tmp .zip]).path
```
agrare, line 154: If you use `Tempfile.create` with a block you don't need to explicitly unlink the temp file, https://ruby-doc.org/stdlib-3.1.1/libdoc/tempfile/rdoc/Tempfile.html#method-c-create

```
Tempfile.create(%w[opentofu-runner-payload .zip]) do |zip_file|
  create_zip_file_from_directory(zip_file.path, template_path)
end
```

And then you don't need the `ensure` https://github.com/ManageIQ/manageiq-providers-embedded_terraform/pull/7/files#diff-6c77e1aede2fb01f9a96887c7b650c186b8daad450c61e621c389b84ffcd1d21R175

lib/terraform/runner.rb
agrare: I think `zip_file_hash` is a little misleading, I expected this to be a e.g. md5/sha2 checksum not the full base64 encoded payload

agrare, line 156: Could you move the Tempfile bits into `create_zip_file_from_directory`?  It doesn't seem like you need the actual file in the rest of this method and you could just return the base64 encoded payload

lib/terraform/runner.rb
```diff
@@ -0,0 +1,293 @@
...
+        )
+        http_response = terraform_runner_client['api/stack/cancel'].post(
+          payload, :content_type => 'application/json'
+        )
+        _log.info("==== Cancel Stack Response: \n #{http_response.body}")
+        Terraform::Runner::Response.parsed_response(http_response)
+      end
+
+      # Wait for TerraformRunner Stack Job to complete
+      def wait_until_completes(stack_id)
```
agrare, line 210: @putmanoj will this method actually be used?  The preferred way of doing this is by using the MiqQueue to check the status and if it isn't completed to requeue itself (aka the Job state machine).

This has the potential to timeout the queue worker if it takes too long.

lib/terraform/runner.rb
jrafanie: In retrospect, this is fine as is.  Ignoring.

lib/terraform/runner.rb
jrafanie: done

lib/terraform/runner.rb
jrafanie: done

lib/terraform/runner.rb
jrafanie: done

lib/terraform/runner.rb
jrafanie: done

lib/terraform/runner.rb
jrafanie: done

jrafanie:
Ok, I pushed a backup of the current branch here: https://github.com/ManageIQ/manageiq-providers-embedded_terraform/compare/master...jrafanie:manageiq-providers-embedded_terraform:terraform-runner-backup

I'm force pushing minor changes.  I've squashed a few "fix" commits.  I've marked the comments as resolved. There's still a few remaining questions I'll try to review and resolve tomorrow.

lib/terraform/runner/response_async.rb
jrafanie: done

lib/terraform/runner.rb
jrafanie: I agree.  @agrare and I chatted about this and we think it makes sense to only have `run_async` and `run` can call `run_async`.  We think `run` (synchronously) was written for ansible because the CLI had that option and it would handle the waiting.  We don't really need this here.  If that's fine, I can remove this and make the changes so `run` is just `run_async`.

jrafanie: let me know what you think @putmanoj

lib/terraform/runner.rb
jrafanie: done

lib/terraform/runner.rb
jrafanie: done

jrafanie:
@putmanoj I added a few commits to use faraday instead of rest-client.  Once you update the docs so I can run the opentofu runner, I'll test it for real.  All the tests pass but I'd like to make sure my changes to the various headers such as ssl verify and authorization work correctly.  I'll need to check the response logic with real responses but I think it's a good start.

lib/terraform/runner.rb
putmanoj: There always will be CREATE action, for starting a Job in terraform-runner  and checking the Job status

jrafanie:
Ok @putmanoj.  I was able to run the simple-template-hcl and run `Terraform::Runner.available?` and `Terraform::Runner.run_async({}, Dir.pwd)` in rails console with the `TERRAFORM_RUNNER_URL` and `TERRAFORM_RUNNER_TOKEN` env variables set and I get the proper API responses with the faraday code in place.  🎉

lib/terraform/runner/response_async.rb
```diff
@@ -0,0 +1,86 @@
...
+      #
+      #   @response
+      # end
+
+      # @return [Terraform::Runner::Response] Response object with all details about the Terraform run, or nil
+      #         if the Terraform is still running
+      def response
+        if running?
+          _log.info("terraform-runner job [#{@stack_id}] is still running ...")
+        end
```
jrafanie, line 50: @putmanoj what's the thinking here?  Is this debugging?  is it needed?

lib/terraform/runner.rb
```diff
@@ -0,0 +1,195 @@
...
+            :ssl => {:verify => verify_ssl}
+          ) do |builder|
+            builder.request(:authorization, 'Bearer', -> { server_token })
+          end
+        end
+      end
+
+      def stack_tenant_id
+        # TODO: fix hardcoded tenant_id
+        'c158d710-d91c-11ed-9fee-d93323035b4e'
```
jrafanie, line 117: @putmanoj What should this be?   Should we fetch it from the ENV?

lib/terraform/runner/response_async.rb
```diff
@@ -0,0 +1,86 @@
...
+
+      # Creates the Terraform::Runner::ResponseAsync object from hash data
+      #
+      # @param hash [Hash] Dumped Terraform::Runner::ResponseAsync object
+      #
+      # @return [Terraform::Runner::ResponseAsync] Terraform::Runner::ResponseAsync Object created from hash data
+      def self.load(hash)
+        # Dump dumps a hash and load accepts a hash, so we must expand the hash to kwargs as new expects kwargs
+        new(**hash)
+      end
```
jrafanie, line 72: Where are we using dump and load?

lib/terraform/runner.rb
```diff
@@ -0,0 +1,194 @@
...
+        }
+        # _log.debug("Payload:>\n, #{payload}")
+
+        http_response = terraform_runner_client.post(
+          "api/stack/create",
+          *json_post_arguments(payload)
+        )
+        _log.debug("==== http_response.body: \n #{http_response.body}")
+        _log.info("stack_job for template: #{template_path} running ...")
+        Terraform::Runner::Response.parsed_response(http_response)
```
jrafanie, line 152: @putmanoj can you test a "successful" run of create_stack_job after my changes?  I'm not sure how to provide credentials and other argument correctly.  Thanks!

jrafanie:
@putmanoj here's a listing of all my changes from your original PR before I started to now:

`git diff origin/terraform-runner-backup putmanoj/terraform-runner`


```diff
diff --git a/lib/terraform/runner.rb b/lib/terraform/runner.rb
index 172fc0a..61b69ab 100644
--- a/lib/terraform/runner.rb
+++ b/lib/terraform/runner.rb
@@ -1,5 +1,4 @@
-require 'rest-client'
-require 'timeout'
+require 'faraday'
 require 'tempfile'
 require 'zip'
 require 'base64'
@@ -10,8 +9,8 @@ module Terraform
       def available?
         return @available if defined?(@available)
 
-        response = terraform_runner_client['api/terraformjobs/count'].get
-        @available = response.code == 200
+        response = terraform_runner_client.get('api/terraformjobs/count')
+        @available = response.status == 200
       rescue
         @available = false
       end
@@ -38,6 +37,10 @@ module Terraform
         Terraform::Runner::ResponseAsync.new(response.stack_id)
       end
 
+      # To simplify clients who may just call run, we alias it to call
+      # run_async.  If we ever need run_sync, we'll need to revisit this.
+      alias run run_async
+
       # Stop running terraform-runner job by stack_id
       #
       # @param stack_id [String] stack_id from the terraforn-runner job
@@ -47,27 +50,6 @@ module Terraform
         cancel_stack_job(stack_id)
       end
 
-      # Runs a template, waits until it terraform-runner job completes, via terraform-runner api
-      #
-      # @param input_vars [Hash] Hash with key/value pairs that will be passed as input variables to the
-      #        terraform-runner run
-      # @param template_path [String] Path to the template we will want to run
-      # @param tags [Hash] Hash with key/values pairs that will be passed as tags to the terraform-runner run
-      # @param credentials [Array] List of Authentication object ids to provide to the terraform run
-      # @param env_vars [Hash] Hash with key/value pairs that will be passed as environment variables to the
-      #        terraform-runner run
-      # @return [Terraform::Runner::Response] Response object with final result of terraform run
-      def run(input_vars, template_path, tags: nil, credentials: [], env_vars: {})
-        _log.debug("Run template: #{template_path}")
-        create_stack_job_and_wait_until_completes(
-          template_path,
-          :input_vars  => input_vars,
-          :tags        => tags,
-          :credentials => credentials,
-          :env_vars    => env_vars
-        )
-      end
-
       # Fetch terraform-runner job result/status by stack_id
       #
       # @param stack_id [String] stack_id from the terraforn-runner job
@@ -83,7 +65,7 @@ module Terraform
       private
 
       def server_url
-        ENV['TERRAFORM_RUNNER_URL'] || 'https://opentofu-runner:27000'
+        ENV.fetch('TERRAFORM_RUNNER_URL', 'https://opentofu-runner:27000')
       end
 
       def server_token
@@ -91,15 +73,11 @@ module Terraform
       end
 
       def stack_job_interval_in_secs
-        ENV['TERRAFORM_RUNNER_STACK_JOB_CHECK_INTERVAL'].to_i
-      rescue
-        10
+        ENV.fetch('TERRAFORM_RUNNER_STACK_JOB_CHECK_INTERVAL', 10).to_i
       end
 
       def stack_job_max_time_in_secs
-        ENV['TERRAFORM_RUNNER_STACK_JOB_MAX_TIME'].to_i
-      rescue
-        120
+        ENV.fetch('TERRAFORM_RUNNER_STACK_JOB_MAX_TIME', 120).to_i
       end
 
       # Create to paramaters as used by terraform-runner api
@@ -108,28 +86,29 @@ module Terraform
       #        terraform-runner run
       # @return [Array] Array of {:name,:value}
       def convert_to_cam_parameters(vars)
-        parameters = []
-        vars&.each do |key, value|
-          parameters.push(
-            {
-              :name  => key,
-              :value => value
-            }
-          )
+        return [] if vars.nil?
+
+        vars.map do |key, value|
+          {
+            :name  => key,
+            :value => value
+          }
         end
-        parameters
       end
 
       # create http client for terraform-runner rest-api
       def terraform_runner_client
-        # TODO: verify ssl
-        verify_ssl = false
-
-        RestClient::Resource.new(
-          server_url,
-          :headers    => {:authorization => "Bearer #{server_token}"},
-          :verify_ssl => verify_ssl
-        )
+        @terraform_runner_client ||= begin
+          # TODO: verify ssl
+          verify_ssl = false
+
+          Faraday.new(
+            :url => server_url,
+            :ssl => {:verify => verify_ssl}
+          ) do |builder|
+            builder.request(:authorization, 'Bearer', -> { server_token })
+          end
+        end
       end
 
       def stack_tenant_id
@@ -137,6 +116,10 @@ module Terraform
         'c158d710-d91c-11ed-9fee-d93323035b4e'
       end
 
+      def json_post_arguments(payload)
+        return JSON.generate(payload), "Content-Type" => "application/json".freeze
+      end
+
       # Create TerraformRunner Stack Job
       def create_stack_job(
         template_path,
@@ -148,45 +131,32 @@ module Terraform
       )
         _log.info("start stack_job for template: #{template_path}")
         tenant_id = stack_tenant_id
-
-        # Temp Zip File
-        # zip_file_path = Tempfile.new(%w/tmp .zip/).path
-        zip_file_path = Tempfile.new(%w[tmp .zip]).path
-        create_zip_file_from_directory(zip_file_path, template_path)
-        zip_file_hash = Base64.encode64(File.binread(zip_file_path))
+        encoded_zip_file = encoded_zip_from_directory(template_path)
 
         # TODO: use tags,env_vars
-        payload = JSON.generate(
-          {
-            :cloud_providers => credentials,
-            :name            => name,
-            :tenantId        => tenant_id,
-            :templateZipFile => zip_file_hash,
-            :parameters      => convert_to_cam_parameters(input_vars)
-          }
-        )
+        payload = {
+          :cloud_providers => credentials,
+          :name            => name,
+          :tenantId        => tenant_id,
+          :templateZipFile => encoded_zip_file,
+          :parameters      => convert_to_cam_parameters(input_vars)
+        }
         # _log.debug("Payload:>\n, #{payload}")
-        http_response = terraform_runner_client['api/stack/create'].post(
-          payload, :content_type => 'application/json'
+
+        http_response = terraform_runner_client.post(
+          "api/stack/create",
+          *json_post_arguments(payload)
         )
         _log.debug("==== http_response.body: \n #{http_response.body}")
         _log.info("stack_job for template: #{template_path} running ...")
         Terraform::Runner::Response.parsed_response(http_response)
-      ensure
-        # cleanup temp zip file
-        FileUtils.rm_rf(zip_file_path) if zip_file_path
-        _log.debug("Deleted #{zip_file_path}")
       end
 
       # Retrieve TerraformRunner Stack Job details
       def retrieve_stack_job(stack_id)
-        payload = JSON.generate(
-          {
-            :stack_id => stack_id
-          }
-        )
-        http_response = terraform_runner_client['api/stack/retrieve'].post(
-          payload, :content_type => 'application/json'
+        http_response = terraform_runner_client.post(
+          "api/stack/retrieve",
+          *json_post_arguments({:stack_id => stack_id})
         )
         _log.info("==== Retrieve Stack Response: \n #{http_response.body}")
         Terraform::Runner::Response.parsed_response(http_response)
@@ -194,99 +164,30 @@ module Terraform
 
       # Cancel/Stop running TerraformRunner Stack Job
       def cancel_stack_job(stack_id)
-        payload = JSON.generate(
-          {
-            :stack_id => stack_id
-          }
-        )
-        http_response = terraform_runner_client['api/stack/cancel'].post(
-          payload, :content_type => 'application/json'
+        http_response = terraform_runner_client.post(
+          "api/stack/cancel",
+          *json_post_arguments({:stack_id => stack_id})
         )
         _log.info("==== Cancel Stack Response: \n #{http_response.body}")
         Terraform::Runner::Response.parsed_response(http_response)
       end
 
-      # Wait for TerraformRunner Stack Job to complete
-      def wait_until_completes(stack_id)
-        interval_in_secs = stack_job_interval_in_secs
-        max_time_in_secs = stack_job_max_time_in_secs
-
-        response = nil
-        Timeout.timeout(max_time_in_secs) do
-          _log.debug("Starting wait for terraform-runner/stack/#{stack_id} completes ...")
-          i = 0
-          loop do
-            _log.debug("loop #{i}")
-            i += 1
-
-            response = retrieve_stack_job(stack_id)
-
-            _log.info("status: #{response.status}")
-
-            case response.status
-            when "SUCCESS"
-              _log.debug("Successful! (stack_job/:#{stack_id})")
-              break
-
-            when "FAILED"
-              _log.info("Failed! (stack_job/:#{stack_id} fails!)")
-              _log.info(response.error_message)
-              break
-
-            when nil
-              _log.info("No status! stack_job/:#{stack_id} must have failed, check response ...")
-              _log.info(response.message)
-              break
-            end
-            _log.info("============\n stack_job/:#{stack_id} status=#{response.status} \n============")
-
-            # sleep interval
-            _log.debug("Sleep for #{interval_in_secs} secs")
-            sleep interval_in_secs
-
-            break unless i < 20
-          end
-          _log.debug("loop ends: ran #{i} times")
-        end
-        response
-      end
-
-      # Create TerraformRunner Stack Job, wait until completes
-      def create_stack_job_and_wait_until_completes(
-        template_path,
-        input_vars: [],
-        tags: nil,
-        credentials: [],
-        env_vars: {},
-        name: "stack-#{rand(36**8).to_s(36)}"
-      )
-        _log.info("create_stack_job_and_wait_until_completes for #{template_path}")
-        response = create_stack_job(
-          template_path,
-          :input_vars  => input_vars,
-          :tags        => tags,
-          :credentials => credentials,
-          :env_vars    => env_vars,
-          :name        => name
-        )
-        wait_until_completes(response.stack_id)
-      end
-
-      # create zip from directory
-      def create_zip_file_from_directory(zip_file_path, template_path)
+      # encode zip of a template directory
+      def encoded_zip_from_directory(template_path)
         dir_path = template_path # directory to be zipped
         dir_path = path[0...-1] if dir_path.end_with?('/')
 
-        _log.debug("Create #{zip_file_path}")
-        Zip::File.open(zip_file_path, Zip::File::CREATE) do |zipfile|
-          Dir.chdir(dir_path)
-          Dir.glob("**/*").reject { |fn| File.directory?(fn) }.each do |file|
-            _log.debug("Adding #{file}")
-            zipfile.add(file.sub("#{dir_path}/", ''), file)
+        Tempfile.create(%w[opentofu-runner-payload .zip]) do |zip_file_path|
+          _log.debug("Create #{zip_file_path}")
+          Zip::File.open(zip_file_path, Zip::File::CREATE) do |zipfile|
+            Dir.chdir(dir_path)
+            Dir.glob("**/*").select { |fn| File.file?(fn) }.each do |file|
+              _log.debug("Adding #{file}")
+              zipfile.add(file.sub("#{dir_path}/", ''), file)
+            end
           end
+          Base64.encode64(File.binread(zip_file_path))
         end
-
-        zip_file_path
       end
     end
   end
diff --git a/lib/terraform/runner/response_async.rb b/lib/terraform/runner/response_async.rb
index 7b62fae..918dbb8 100644
--- a/lib/terraform/runner/response_async.rb
+++ b/lib/terraform/runner/response_async.rb
@@ -74,16 +74,12 @@ module Terraform
       private
 
       def completed?(status)
-        # IF NOT SUCCESS,FAILED,CANCELLED
-        if status
-          return (
-            status.start_with?("SUCCESS", "FAILED") ||
-            # @response.status == "ERROR" ||
-            status == "CANCELLED"
-          )
+        case status.to_s.upcase
+        when "SUCCESS", "FAILED", "CANCELLED"
+          true
+        else
+          false
         end
-
-        false
       end
     end
   end
diff --git a/spec/lib/terraform/runner_spec.rb b/spec/lib/terraform/runner_spec.rb
index 2b23008..49ab330 100644
--- a/spec/lib/terraform/runner_spec.rb
+++ b/spec/lib/terraform/runner_spec.rb
@@ -21,89 +21,6 @@ RSpec.describe(Terraform::Runner) do
     end
   end
 
-  describe ".run hello-world with no input vars ( nil argument )" do
-    let(:input_vars) { nil }
-
-    before do
-      ENV["TERRAFORM_RUNNER_URL"] = "https://1.2.3.4:7000"
-
-      stub_request(:post, "https://1.2.3.4:7000/api/stack/create")
-        .with(:body => hash_including({:parameters => []}))
-        .to_return(
-          :status => 200,
-          :body   => @hello_world_create_response.to_json
-        )
-
-      stub_request(:post, "https://1.2.3.4:7000/api/stack/retrieve")
-        .with(:body => hash_including({:stack_id => @hello_world_retrieve_response['stack_id']}))
-        .to_return(
-          :status => 200,
-          :body   => @hello_world_retrieve_response.to_json
-        )
-    end
-
-    it "runs a hello-world terraform template" do
-      response = Terraform::Runner.run(input_vars, File.join(__dir__, "runner/data/hello-world"))
-
-      expect(response.status).to(eq('SUCCESS'), "terraform-runner failed with:\n#{response.status}")
-      expect(response.message).to(include('greeting = "Hello World"'))
-      expect(response.stack_id).to(eq(@hello_world_retrieve_response['stack_id']))
-      expect(response.action).to(eq('CREATE'))
-      expect(response.stack_name).to(eq(@hello_world_retrieve_response['stack_name']))
-      expect(response.details.keys).to(eq(%w[resources outputs]))
-    end
-  end
-
-  describe ".run hello-world with input_vars" do
-    let(:input_vars) { {:name => 'Mumbai'} }
-
-    def verify_request_and_respond(request)
-      body = JSON.parse(request.body)
-
-      # verify parameters
-      expect(body['parameters'].length).to(eq(1))
-      data = body['parameters'][0]
-      expect(data['name']).to(eq('name'))
-      expect(data['value']).to(eq('Mumbai'))
-
-      # verify other attributes
-      expect(body['name']).not_to(be_empty)
-      expect(body['tenantId']).not_to(be_empty)
-      expect(body['templateZipFile']).not_to(be_empty)
-
-      @hello_world_create_response.to_json
-    end
-
-    before do
-      ENV["TERRAFORM_RUNNER_URL"] = "https://1.2.3.4:7000"
-
-      stub_request(:post, "https://1.2.3.4:7000/api/stack/create")
-        .to_return(->(request) { {:body => verify_request_and_respond(request)} })
-
-      response = @hello_world_retrieve_response.clone
-      response['message'] = response['message'].gsub('World', 'Mumbai')
-      response['details']['outputs'][0]['value'] = response['details']['outputs'][0]['value'].sub('World', 'Mumbai')
-
-      stub_request(:post, "https://1.2.3.4:7000/api/stack/retrieve")
-        .with(:body => hash_including({:stack_id => @hello_world_retrieve_response['stack_id']}))
-        .to_return(
-          :status => 200,
-          :body   => response.to_json
-        )
-    end
-
-    it "runs a hello-world terraform template" do
-      response = Terraform::Runner.run(input_vars, File.join(__dir__, "runner/data/hello-world"))
-
-      expect(response.status).to(eq('SUCCESS'), "terraform-runner failed with:\n#{response.status}")
-      expect(response.message).to(include('greeting = "Hello Mumbai"'))
-      expect(response.stack_id).to(eq(@hello_world_retrieve_response['stack_id']))
-      expect(response.action).to(eq('CREATE'))
-      expect(response.stack_name).to(eq(@hello_world_retrieve_response['stack_name']))
-      expect(response.details.keys).to(eq(%w[resources outputs]))
-    end
-  end
-
   context '.run_async hello-world' do
     describe '.run_async' do
       create_stub = nil
@@ -142,7 +59,10 @@ RSpec.describe(Terraform::Runner) do
         expect(response.stack_name).to(eq(@hello_world_create_response['stack_name']))
         expect(response.message).to(be_nil)
         expect(response.details).to(be_nil)
+      end
 
+      it "is aliased as run" do
+        expect(Terraform::Runner.method(:run)).to eq(Terraform::Runner.method(:run_async))
       end
     end
```

lib/terraform/runner.rb
jrafanie: I've removed run being run synchronously and now it's an alias of run_async.  Let me know if I didn't do it right.

lib/terraform/runner/response_async.rb
agrare: This looks like a copy&paste from Ansible::Runner::ResponseAsync because that is interacting with the filesystem to get the response.  No need here if it isn't used.

lib/terraform/runner.rb
```diff
@@ -0,0 +1,193 @@
...
+          :templateZipFile => encoded_zip_file,
+          :parameters      => convert_to_cam_parameters(input_vars)
+        }
+        # _log.debug("Payload:>\n, #{payload}")
+
+        http_response = terraform_runner_client.post(
+          "api/stack/create",
+          *json_post_arguments(payload)
+        )
+        _log.debug("==== http_response.body: \n #{http_response.body}")
```
agrare, line 149: Not for this PR, but we have a plugin logger (`$embedded_terraform_log`) which is typically used for api client logging we could probably use that when we're just dumping the request / response to the logs

lib/terraform/runner.rb
jrafanie: changed hardcoded values to all 0s for now

lib/terraform/runner/response_async.rb
jrafanie: dropped this for now, we can add it if we need to serialize these objects

lib/terraform/runner/response_async.rb
```diff
@@ -0,0 +1,67 @@
...
+        raise "No job running to stop" if !running?
+
+        Terraform::Runner.stop_async(@stack_id)
+      end
+
+      # Re-Fetch async job's response
+      def refresh_response
+        @response = Terraform::Runner.fetch_result_by_stack_id(@stack_id)
+
+        @response
```
agrare, line 34: Really minor but the "return" of an assignment is the value anyway so you don't need this
```suggestion
```

lib/terraform/runner/response_async.rb
agrare: Yeah we could probably drop this method entirely and just go with an `attr_reader :response`

lib/terraform/runner/response_async.rb
```diff
@@ -0,0 +1,67 @@
...
+      # Response object designed for holding full response from terraform-runner stack job
+      #
+      # @param stack_id [String] terraform-runner stack job - stack_id
+      def initialize(stack_id)
+        @stack_id = stack_id
+      end
+
+      # @return [Boolean] true if the terraform stack job is still running, false when it's finished
+      def running?
+        return false if @response && completed?(@response.status)
```
agrare, line 15: You could do `completed?(@response&.status)` since `completed?` would handle a nil and return false

agrare (approved):
Couple of minor follow-up comments but overall looks good

lib/terraform/runner.rb
```diff
@@ -0,0 +1,193 @@
...
+require 'zip'
+require 'base64'
+
+module Terraform
+  class Runner
+    class << self
+      def available?
+        return @available if defined?(@available)
+
+        response = terraform_runner_client.get('api/terraformjobs/count')
```
Fryguy, line 12: Super-tiny-minor, but do we really need a list of all jobs to show availability?  That is, is there
a) a simpler query like maybe just hitting /api itself?
b) can we use a HEAD request?

jrafanie: how about ping?

```ruby
irb(main):003:0> Terraform::Runner.send(:terraform_runner_client).get('api/ping').status
=> 200
```

jrafanie:
I'll tackle the ping endpoint and other comments in a followup PR

lib/terraform/runner.rb
jrafanie: opened https://github.com/ManageIQ/manageiq-providers-embedded_terraform/pull/14 for this

PR #11, opened 2024-04-01
Update codeclimate channel to the latest in manageiq-style

agrare:
Update codeclimate channel to the latest in manageiq-style


agrare:
This repo was missed because we didn't have it in manageiq-release yet.

PR #12, opened 2024-04-10
Add EmbeddedTerraform Stack and Job

agrare:
Adds a `EmbeddedTerraform::AutomationManager::Stack` which is an `OrchestrationStack` subclass
Adds `EmbeddedTerraform::AutomationManager::Template#run` which is called by `EmbeddedTerraform::AutomationManager::Stack.create_stack` and creates&starts a `Job` state-machine
Adds `EmbeddedTerraform::AutomationManager::Job` which is a `::Job` state-machine and is responsible for running and polling for completion any terraform runner tasks.

### DIFF
--- a/app/models/manageiq/providers/embedded_terraform/automation_manager/job.rb
+++ b/app/models/manageiq/providers/embedded_terraform/automation_manager/job.rb
@@ -1,14 +1,51 @@
 class ManageIQ::Providers::EmbeddedTerraform::AutomationManager::Job < Job
+  def self.create_job(template, env_vars, input_vars, credentials, poll_interval: 1.minute)
+    super(
+      :template_id   => template.id,
+      :env_vars      => env_vars,
+      :input_vars    => input_vars,
+      :credentials   => credentials,
+      :poll_interval => poll_interval,
+    )
+  end
+
   def start
-    queue_signal(:poll_runner)
+    queue_signal(:pre_execute)
+  end
+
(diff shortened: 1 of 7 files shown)


agrare:
Okay @jrafanie this is ready for review

app/models/manageiq/providers/embedded_terraform/automation_manager/job.rb
```diff
@@ -0,0 +1,129 @@
...
+      :post_execute => {'running'          => 'post_execute'},
+      :finish       => {'*'                => 'finished'},
+      :abort_job    => {'*'                => 'aborting'},
+      :cancel       => {'*'                => 'canceling'},
+      :error        => {'*'                => '*'}
+    }
+  end
+
+  def poll_interval
+    options.fetch(:poll_interval, 1.minute).to_i
```
jrafanie, line 85: 1 minute seems like a long time.  How long are we expecting short and long templates to take to run?  Would 30 seconds be better?

agrare: Probably a good question for @putmanoj

jrafanie (approved):
Looks good to me, we can change that poll interval in a followup.

jrafanie:
@putmanoj I think this reworks some of the POC code you had around workflows, I'm merging today.  Let me know if you need help getting your workflows branch working with this.

PR #3, opened 2024-03-04
Add EmbeddedTerraform UI

agrare:
Depends on:
- [x] https://github.com/ManageIQ/manageiq/pull/22995
- [ ] https://github.com/ManageIQ/manageiq-ui-classic/pull/9117

Depend
TODO
- [x] Routes
- [x] Menu Items
- [x] `db_to_controller`
- [x] `title_from_layout`
- [x] `WorkflowRepositoryForm`, `WorkflowCredentialsForm` javascript forms

Menu:
image

Credentials Show List:
image

Credentials Show:
image

Credentials Show -> Repositories Show List:
image

Repositories Show List:
image

Repositories Show:
image

Repositories Show -> Templates Show List:
image

Templates Show List:
image

Templates Show:
image

PR #16, opened 2024-04-22
Add cloud creds support for terraform runner

putmanoj:
* Add cloud credentials classes for embedded-terraform with DDF configuration
* Add cloud credentails classes with authentication connection parameters for terraform-runner


app/models/manageiq/providers/embedded_terraform/automation_manager/scm_credential.rb
```diff
@@ -1,5 +1,5 @@
 class ManageIQ::Providers::EmbeddedTerraform::AutomationManager::ScmCredential < ManageIQ::Providers::EmbeddedTerraform::AutomationManager::Credential
   include ManageIQ::Providers::EmbeddedAutomationManager::ScmCredentialMixin
 
-  FRIENDLY_NAME = "Embedded Terraform SCM Credential".freeze
+  FRIENDLY_NAME = "Embedded Terraform Credential".freeze
```
agrare, line 4: Why remove this?  I think it is helpful to differentiate this from the other terraform credential types.

putmanoj: It was simpler removing 'one' word,  then fixing 12 failures, only for one word :-|  ...
will take look tomorrow  ...

agrare: What failures?  This is just for display I thought

putmanoj: might be something silly,  only did not want put too much time on it, at the time ...

putmanoj: it uses 'shared_examples'  https://github.com/putmanoj/manageiq-providers-embedded_terraform/blob/add-cloud-creds-support-for-terraform-runner/spec/models/manageiq/providers/embedded_terraform/automation_manager/credential_spec.rb#L24

app/models/manageiq/providers/embedded_terraform/automation_manager/ibmcloud_credential.rb
```diff
@@ -0,0 +1,58 @@
+class ManageIQ::Providers::EmbeddedTerraform::AutomationManager::IbmcloudCredential < ManageIQ::Providers::EmbeddedTerraform::AutomationManager::CloudCredential
```
jrafanie, line 1: The naming of the file and casing of the class doesn't look right.

I suggest we follow what we use elsewhere:

https://github.com/ManageIQ/manageiq-providers-ibm_cloud/blob/master/app/models/manageiq/providers/ibm_cloud/container_manager.rb

`class ManageIQ::Providers::IbmCloud::ContainerManager < ManageIQ::Providers::Kubernetes::ContainerManager`

in file: `providers/ibm_cloud/container_manager.rb`

So, `IbmCloudCredential` in `automation_manager/ibm_cloud_credential.rb`

app/models/manageiq/providers/embedded_terraform/automation_manager/scm_credential.rb
agrare, line 4:
```suggestion
  FRIENDLY_NAME = "Embedded Terraform SCM Credential".freeze
```

spec/models/manageiq/providers/embedded_terraform/automation_manager/credential_spec.rb
```diff
@@ -0,0 +1,587 @@
...
+        expect(record).to(be_a(credential_class))
+        expect(record.manager).to(eq(manager))
+        expected_values.each do |attr, val|
+          expect(record.send(attr)).to(eq(val))
+        end
+      end
+
+      it ".create_in_provider_queue queues a create task" do
+        task_id       = credential_class.create_in_provider_queue(manager.id, params)
+        expected_name = "Creating #{described_class::FRIENDLY_NAME} (name=#{params[:name]})"
```
agrare, line 65:
```suggestion
        expected_name = "Creating #{credential_class::FRIENDLY_NAME} (name=#{params[:name]})"
```

spec/models/manageiq/providers/embedded_terraform/automation_manager/credential_spec.rb
```diff
@@ -0,0 +1,587 @@
...
+
+        # Doesn't muck up old attrs
+        previous_params_to_attrs.each do |attr, value|
+          expect(result.send(attr)).to(eq(value))
+        end
+      end
+
+      it "#update_in_provider_queue" do
+        task_id   = terraform_cred.update_in_provider_queue(update_params)
+        task_name = "Updating #{described_class::FRIENDLY_NAME} (name=#{terraform_cred.name})"
```
agrare, line 108:
```suggestion
        task_name = "Updating #{credential_class::FRIENDLY_NAME} (name=#{terraform_cred.name})"
```

spec/models/manageiq/providers/embedded_terraform/automation_manager/credential_spec.rb
```diff
@@ -0,0 +1,587 @@
...
+      let(:terraform_cred) { credential_class.raw_create_in_provider(manager, params) }
+
+      it "#delete_in_provider will delete the record" do
+        expect(Notification).to(receive(:create!).never)
+        terraform_cred.delete_in_provider
+      end
+
+      it "#delete_in_provider_queue will queue a a delete task" do
+        task_id   = terraform_cred.delete_in_provider_queue
+        task_name = "Deleting #{described_class::FRIENDLY_NAME} (name=#{terraform_cred.name})"
```
agrare, line 135:
```suggestion
        task_name = "Deleting #{credential_class::FRIENDLY_NAME} (name=#{terraform_cred.name})"
```

agrare (changes requested):

app/models/manageiq/providers/embedded_terraform/automation_manager/cloud_credential.rb
```diff
@@ -0,0 +1,2 @@
+class ManageIQ::Providers::EmbeddedTerraform::AutomationManager::CloudCredential < ManageIQ::Providers::EmbeddedTerraform::AutomationManager::Credential
+end
```
agrare, line 2: @putmanoj what does the CloudCredential intermediate class buy us here?  It seems more confusing because VsphereCredential inherits from it but vSphere is an Infra manager not a Cloud manager.

app/models/manageiq/providers/embedded_terraform/automation_manager/amazon_credential.rb
```diff
@@ -0,0 +1,68 @@
...
+
+  alias security_token auth_key
+
+  def self.display_name(number = 1)
+    n_('Credential (Amazon)', 'Credentials (Amazon)', number)
+  end
+
+  def self.params_to_attributes(params)
+    attrs = super.dup
+    attrs[:auth_key] = attrs.delete(:security_token) if attrs.key?(:security_token)
```
agrare, line 61: Why call a property security_token if we're just going to store it in auth_key, could you use `"auth_key"` instead up here :point_up: https://github.com/ManageIQ/manageiq-providers-embedded_terraform/pull/16/files#diff-10eac92f3aa8adb511c059784672e62020e8e6172d8154c295c6a5678c75007fR29-R30

app/models/manageiq/providers/embedded_terraform/automation_manager/google_credential.rb
```diff
@@ -0,0 +1,71 @@
...
+  alias ssh_key_data auth_key
+
+  def self.display_name(number = 1)
+    n_('Credential (Google)', 'Credentials (Google)', number)
+  end
+
+  def self.params_to_attributes(params)
+    attrs = super.dup
+
+    attrs[:auth_key] = attrs.delete(:ssh_key_data) if attrs.key?(:ssh_key_data)
```
agrare, line 62: Basically the same question here as above (and looks like for a number of these).  Maybe there is a good reason? IDK

app/models/manageiq/providers/embedded_terraform/automation_manager/openstack_credential.rb
```diff
@@ -0,0 +1,131 @@
...
+      attrs[:options][:region]  = attrs.delete(:region)  if attrs.key?(:region)
+
+      attrs[:options][:cacert]     = attrs.delete(:cacert)     if attrs.key?(:cacert)
+      attrs[:options][:clientcert] = attrs.delete(:clientcert) if attrs.key?(:clientcert)
+      attrs[:options][:clientkey]  = attrs.delete(:clientkey)  if attrs.key?(:clientkey)
+
+      attrs[:options][:insecure] = attrs.delete(:insecure) if attrs.key?(:insecure)
+    end
+
+    attrs
```
jrafanie, line 117: I think something like this does what you're trying to do:

```suggestion
    if attrs.keys & %i[host domain project region cacert clientcert clientkey insecure]
      attrs.slice!(:host, :domain, :project, :region, :cacert, :clientkey, :insecure)
    end

    {:options => attrs}
```

jrafanie: Are there other values you're concerned about retaining?  The slice option will only remove all key/values that aren't in the list so if you have others that aren't in that list, they'll be lost.

agrare: I think `attrs` will have more than just these keys so we can't replace the top level hash with just `{:options =>`
What about something like

```
options = super.dup
attrs = options.slice!(:host, :domain, :project, :region, :cacert, :clientkey, :insecure)
attrs.merge!(:options => options) unless options.blank?
attrs
```

`attrs` will now have the rest of the keys that aren't in the `slice!` which should be the rest of the top level keys.

app/models/manageiq/providers/embedded_terraform/automation_manager/vsphere_credential.rb
```diff
@@ -0,0 +1,58 @@
+class ManageIQ::Providers::EmbeddedTerraform::AutomationManager::VsphereCredential < ManageIQ::Providers::EmbeddedTerraform::AutomationManager::CloudCredential
```
jrafanie, line 1: @agrare thoughts on the casing on `Vsphere`?  I see both kinds, 🤣 

`VSphereAutomation` gem used in the vmware provider and `VsphereProvisioning`, `VsphereReplication`, etc. in intersight client, oci.

agrare: Technically the name is `vSphere` so if we are capitalizing the first letter `VSphere` would be most accurate, but honestly I think `Vsphere` is fine here

agrare: Though I think "wrong answers only" for vMwAre VsPhere would be pretty funny

jrafanie: nerd snipe Adam right before he has to start a call.  🏆 

Oops.  haha, I forgot it's actually vSphere.  Ok, so Vsphere is fine.

agrare: Actually it was https://github.com/ManageIQ/manageiq-providers-embedded_terraform/pull/16#discussion_r1576298959 that did it :D

app/models/manageiq/providers/embedded_terraform/automation_manager/openstack_credential.rb
```diff
@@ -0,0 +1,131 @@
...
+      :helperText     => N_('The client certificate for SSL client authentication.'),
+      :componentClass => 'textarea',
+      :name           => 'clientcert',
+      :id             => 'clientcert',
+      :type           => 'password',
+    },
+    {
+      :component  => 'checkbox',
+      :label      => N_('Trust self-signed SSL certificates ?'),
+      :helperText => N_('Trust self-signed SSL certificates ?'),
```
jrafanie, line 82: Is this the same thing as VERIFY_NONE (when insecure true) vs. VERIFY_PEER (when insecure false)?  We call that `SSL verification` or `SSL Verify Mode` elsewhere.

app/models/manageiq/providers/embedded_terraform/automation_manager/amazon_credential.rb
jrafanie: Good question. We also have an alias for security_token from auth_key so hopefully we can use auth_key everywhere to keep it consistent.

app/models/manageiq/providers/embedded_terraform/automation_manager/openstack_credential.rb
jrafanie: Note, we have this pattern in `params_to_attributes` to pull out specific keys from `attrs` and store them in the `:options` subkey instead.  Perhaps extract a method `move_keys_to_options` from above.

jrafanie: Or not... this one is the most complicated one.

lib/terraform/runner.rb
```diff
@@ -119,6 +103,19 @@ def json_post_arguments(payload)
...
+        cloud_providers = []
+        credentials.each do |cred|
+          runner_cred = Terraform::Runner::Credential.new(cred.id)
+          cloud_providers.push(
+            {
+              'connection_parameters' => runner_cred.connection_parameters
+            }
+          )
+        end
+        cloud_providers
```
jrafanie, line 116:
```suggestion
        credentials.collect do |cred|
            {
              'connection_parameters' => Terraform::Runner::Credential.new(cred.id).connection_parameters
            }
        end
```

lib/terraform/runner.rb
```diff
@@ -119,6 +103,19 @@ def json_post_arguments(payload)
         return JSON.generate(payload), "Content-Type" => "application/json".freeze
       end
 
+      def to_cloud_providers_conn_params(credentials)
```
jrafanie, line 106: how about `provider_connection_parameters`... we don't really care that they're clouds, just that the we'll ask the instance of the class to build the connection parameters.

app/models/manageiq/providers/embedded_terraform/automation_manager/amazon_credential.rb
agrare: Yeah I understand having the `:label` be something specific but wasn't sure what having the internal key being different from the column bought us

lib/terraform/runner/credential.rb
```diff
@@ -0,0 +1,31 @@
+module Terraform
+  class Runner
+    class Credential
+      attr_reader :auth
+
+      def self.new(authentication_id)
```
jrafanie, line 6: Suggest we use `find` so it looks like active record find where you pass an id.  It doesn't matter if it creates a new object or find an existing one, we only care that it returns an instance of the proper subclass object.

lib/terraform/runner/ibmcloud_credential.rb
```diff
@@ -0,0 +1,25 @@
+module Terraform
+  class Runner
+    class IbmcloudCredential < Credential
+      def self.auth_type
+        "ManageIQ::Providers::EmbeddedTerraform::AutomationManager::IbmcloudCredential"
```
jrafanie, line 5: same as above, suggest IbmCloudCredential

app/models/manageiq/providers/embedded_terraform/automation_manager/vsphere_credential.rb
jrafanie: I knew if I did it halfway, you'd finish it

app/models/manageiq/providers/embedded_terraform/automation_manager/openstack_credential.rb
agrare: > Note, we have this pattern in params_to_attributes to pull out specific keys from attrs and store them in the :options subkey instead. Perhaps extract a method move_keys_to_options from above.

Yeah I wonder if maybe defining accessors would help clean this up?  Like

```
def clientkey
  options && options[:clientkey]
end

def clientkey=(val)
  options ||= {}
  options[:clientkey] = val
end
```

Then we probably don't need to override `params_to_attributes` at all.  This would be a lot of change maybe something for a followup

app/models/manageiq/providers/embedded_terraform/automation_manager/azure_credential.rb
```diff
@@ -0,0 +1,80 @@
...
+
+    attrs
+  end
+
+  def client
+    options && options[:client]
+  end
+
+  def tenant
+    options && options[:tenant]
```
jrafanie, line 74: I know this is in connection to the azure tenant, but since we also have a tenant concept in manageiq, I wonder if it makes sense to be explicit that this is the cloud_provider_tenant.  In this case, I think it's a name and not an id/number.

lib/terraform/runner.rb
```diff
@@ -134,11 +131,11 @@ def create_stack_job(
 
         # TODO: use tags,env_vars
         payload = {
-          :cloud_providers => credentials,
+          :cloud_providers => to_cloud_providers_conn_params(credentials),
```
jrafanie, line 134: See tenant comment above.  I wonder if tenant_id should be provider_tenant_name if it's a name or provider_tenant_id if it's a number.  We have tenant_id columns on many tables so this could be easily confused.

app/models/manageiq/providers/embedded_terraform/automation_manager/azure_credential.rb
agrare: Yeah that's a good point, in the provider code we call this a "cloud_tenant"

jrafanie:
With https://github.com/ManageIQ/manageiq-providers-embedded_terraform/pull/3 and https://github.com/ManageIQ/manageiq-ui-classic/pull/9117 locally and this PR... here are some screenshots of the credentials we're defining here:

amazon:
image

azure:
image

google:
image

ibm cloud (quite wordy)
image

openstack 1:
image

openstack 2:

image

scm:
image

vsphere:
image

app/models/manageiq/providers/embedded_terraform/automation_manager/cloud_credential.rb
putmanoj: Conceptually you may be right, but in implementation it just one of the (cloud) providers,  it will show up in UI for selection, as one of the types supported with AWS,GCP, etc  

* so also in ansible you have vmware inheriting from CloudCredential‎ https://github.com/ManageIQ/manageiq/blob/master/app/models/manageiq/providers/embedded_ansible/automation_manager/vmware_credential.rb#L1   
  ( renamed as `vsphere` for terraform, as provider is also named as such  https://github.com/hashicorp/terraform-provider-vsphere ) 

* similarly in CAM

app/models/manageiq/providers/embedded_terraform/automation_manager/amazon_credential.rb
putmanoj: Just followed the pattern of what is in ansible implemetation - just a copy the same 
https://github.com/ManageIQ/manageiq/blob/master/app/models/manageiq/providers/embedded_ansible/automation_manager/amazon_credential.rb#L52

app/models/manageiq/providers/embedded_terraform/automation_manager/google_credential.rb
putmanoj: Again same pattern as in ansible implementation ...

https://github.com/ManageIQ/manageiq/blob/master/app/models/manageiq/providers/embedded_ansible/automation_manager/google_credential.rb#L54

app/models/manageiq/providers/embedded_terraform/automation_manager/google_credential.rb
```diff
@@ -0,0 +1,71 @@
...
+  API_OPTIONS = {
+    :type       => 'cloud',
+    :label      => N_('Google Compute Engine'),
+    :attributes => API_ATTRIBUTES
+  }.freeze
+
+  alias ssh_key_data auth_key
+
+  def self.display_name(number = 1)
+    n_('Credential (Google)', 'Credentials (Google)', number)
```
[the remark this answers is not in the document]
putmanoj: As just got started with ruby/rails/manageiq ..., still some nuances elude me,   

simply taken from ansible implementation https://github.com/ManageIQ/manageiq/blob/master/app/models/manageiq/providers/embedded_ansible/automation_manager/google_credential.rb#L41C20-L41C22

will have to go by your suggestion, what is needed ?

app/models/manageiq/providers/embedded_terraform/automation_manager/openstack_credential.rb
putmanoj: Yes,  we can improve the label text ,

putmanoj: Though by default `insecure` is false by default, so is OFF or 'unchecked'. 
https://registry.terraform.io/providers/terraform-provider-openstack/openstack/latest/docs#insecure

 If we use `Verify SSL`  as label, then meaning changes, ...   
 the checkbox needs to be 'ON' or 'checked' by default.  Just don't know enough DDF yet, to know how to make checkbox 'checked'  by default.

putmanoj: @jrafanie ,  any other label suggestion ?

app/models/manageiq/providers/embedded_terraform/automation_manager/azure_credential.rb
putmanoj: Again this was taken from ansible implementation -> https://github.com/ManageIQ/manageiq/blob/master/app/models/manageiq/providers/embedded_ansible/automation_manager/azure_credential.rb#L89

The azure 'tenant' is available in `options`.
Will remove `def tenant` ,  can't say about ansible,  who needs separate method for terraform anyways ?

putmanoj: Anyways, I replaced 'tenant' with `provider_tenant`

app/models/manageiq/providers/embedded_terraform/automation_manager/amazon_credential.rb
```diff
@@ -0,0 +1,68 @@
...
+      :helperText => N_('Security Token Service(STS) Token for this credential'),
+      :name       => 'security_token',
+      :id         => 'security_token',
+      :type       => 'password',
+      :maxLength  => 1024
+    },
+    {
+      :component  => 'text-field',
+      :label      => N_('Region'),
+      :helperText => N_('AWS Region where the provider will operate. The Region must be set.'),
```
jrafanie, line 37: Do we need to say "The region must be set."? I don't see this for other `isRequired => true` fields.  We also highlight the field as required if you don't provide a value.

app/models/manageiq/providers/embedded_terraform/automation_manager/openstack_credential.rb
jrafanie: I agree, we should use positive terms "Verify SSL" with it checked/enabled by default.  We can with @GilbertCherrie if we're not seeing it "on" by default.

spec/support/evm_spec_helper.rb
```diff
@@ -0,0 +1,25 @@
...
+    MiqRegion.seed
+    miq_server ||= local_miq_server
+    role = ServerRole.find_by(:name => role_name) || FactoryBot.create(:server_role, :name => role_name, **create_options)
+    miq_server.assign_role(role).update(:active => true)
+  end
+
+  def self.assign_embedded_terraform_role(miq_server = nil)
+    assign_role("embedded_terraform", :miq_server => miq_server, :create_options => {:max_concurrent => 0})
+  end
+end
```
jrafanie, line 25: I think this whole file can be removed.  assign_role should be available to us from the repo as we're a plugin:  https://github.com/ManageIQ/manageiq/blob/e9dbcc138963b1072d82eb73538ecc2f53d70779/spec/support/evm_spec_helper.rb#L24-L29

jrafanie (commented):
@putmanoj let ms know if there's any suggested changes you'd like me to work on... Some of these can be things we do in followup PRs.  Anything resolved, we can mark resolved and any still remaining, we can choose to not resolve at all or leave them for followup work.

lib/terraform/runner/credential.rb
putmanoj: Again used the same pattern, as in ansible implementation, https://github.com/ManageIQ/manageiq/blob/master/lib/ansible/runner/credential.rb#L7

Yes it uses `find` but I suppose, `new` also makes sense, as creates a new instance of `Terraform::Runner::Credential`

lib/terraform/runner.rb
putmanoj: renamed the `tenant` field as `provider_tenant`

lib/terraform/runner/credential.rb
jrafanie: Yeah, at second thought, this is like a builder/factory method.  It's fine but generally `new` creates instances of the class whereas a builder/factory could build new objects based on the inputs and they don't need to be the same class.  We can keep it if ansible did it this way, but I try to avoid overriding new.

app/models/manageiq/providers/embedded_terraform/automation_manager/ibm_cloud_credential.rb
```diff
@@ -0,0 +1,58 @@
...
+      :label      => N_('IBM Cloud Classic Infrastructure User Name'),
+      :helperText => N_('The User Name for IBM Cloud Classic Infrastructure. A value for this field is required when using classic IBM Cloud resources. A valid connection must have value for IBM Cloud API Key or, IBM Cloud Classic Infrastructure User Name and IBM Cloud Classic Infrastructure API Key.'),
+      :name       => 'classic_user',
+      :id         => 'classic_user',
+      :maxLength  => 100,
+    },
+    {
+      :component  => 'password-field',
+      :label      => N_('IBM Cloud Classic Infrastructure API Key'),
+      :helperText => N_('The API key for IBM Cloud Classic Infrastructure A value for this field is required when using classic IBM Cloud resources. A valid connection must have value for IBM Cloud API Key or, IBM Cloud Classic Infrastructure User Name and IBM Cloud Classic Infrastructure API Key.'),
```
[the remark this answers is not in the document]
jrafanie: image


There's a lot of repeated phrases with one or two words changed.  I think it's easy to miss the point:

You need to provide an IBM Cloud API Key or a IBM Cloud Classic Infra User and a Classic Infra API Key.

app/models/manageiq/providers/embedded_terraform/automation_manager/google_credential.rb
agrare: Right, I don't think we should blindly copy&paste from other code without thinking about it first :)

app/models/manageiq/providers/embedded_terraform/automation_manager/amazon_credential.rb
agrare: @putmanoj I see you resolved this comment but didn't change the code.  I agree with @jrafanie saying `The Region must be set` is redundant when we already have the field required.

app/models/manageiq/providers/embedded_terraform/automation_manager/amazon_credential.rb
agrare: This can be done in a follow-up

jrafanie: yup, I'll do this in a follow up. no need to delay merging on this.

app/models/manageiq/providers/embedded_terraform/automation_manager/google_credential.rb
jrafanie: I'll get this for followup.

app/models/manageiq/providers/embedded_terraform/automation_manager/openstack_credential.rb
jrafanie: yup, followup is good

app/models/manageiq/providers/embedded_terraform/automation_manager/openstack_credential.rb
jrafanie: We can merge as is and I'll work with @GilbertCherrie if I can't figure out how to correctly switch the logic.  I like "verify ssl" true/false as insecure "false" is a double negative and hard to understand.

app/models/manageiq/providers/embedded_terraform/automation_manager/ibm_cloud_credential.rb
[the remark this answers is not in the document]
jrafanie: We can simplify this in a followup.

app/models/manageiq/providers/embedded_terraform/automation_manager/cloud_credential.rb
putmanoj: It is not about copy&paste,  ...   question is - did not want a different experience for Terraform ?

-  In CAM, the `VMware vSphere` is implemented as one of the (cloud) providers, so as per CAM User point of view, did not want it be any different, in CAM it is one of Cloud crediential. 
- Also in ServiceCatalog UI,  for  Embedded-Ansible/Paybook,  I can see `VMware` as one of the Cloud Types (not different from CAM).

So, are we going to have different user experience in UI,  for `VMware vSphere` vs Amazon in Service-Catalog ? Are we going show `VMware vSphere`  in different dropdown ?
 
Screenshot from 2024-04-29 16-23-34

putmanoj: Perhaps,  we need `CloudCredential` to distinguish from  `ScmCerdential`.   

So both  `CloudCredential` & `InfraCredential`(if we create intermediate)  will need to have same parent, something like `ProviderCredential`.

agrare: On the workflows side we have a WorkflowCredential that indicates a cred that is used by a workflow.
I think we should have a TemplateCredential that all of these which are intended to be used by templates should inherit from

jrafanie: Would a `TemplateCredential` be different from `ManageIQ::Providers::EmbeddedTerraform::AutomationManager::Credential`?  I thought they'd be the same.  If they're different, then, yes, we could use it and then implement cloud vs. infra creds if we needed that intermediate class.

agrare: > Would a TemplateCredential be different from ManageIQ::Providers::EmbeddedTerraform::AutomationManager::Credential? I thought they'd be the same.

I think the point is to differentiate from an ScmCredential which `< Credential` as well.

> and then implement cloud vs. infra creds if we needed that intermediate class.

I believe "cloud credentials" is naming coming from ansible tower and not something we need to maintain here

putmanoj: Also in IA/CAM, the naming convention is  "cloud credentails" (nothing to do with Ansible tower).  
All the supported ones are cloud-credentials,  only `VMware` might be the odd one.

Though `TemplateCredential`  or perhaps `TerraformCredential`  are more meaningful.      They then can be different from `ScmCredential`.

jrafanie: We decided to do this in a followup PR so we can get this merged as is:

Summarizing the comments:

Template subclasses credential:

```ruby
class ManageIQ::Providers::EmbeddedTerraform::AutomationManager::TemplateCredential < 
ManageIQ::Providers::EmbeddedTerraform::AutomationManager::Credential
```

scm credential is unchanged:

```ruby
class ManageIQ::Providers::EmbeddedTerraform::AutomationManager::ScmCredential < ManageIQ::Providers::EmbeddedTerraform::AutomationManager::Credential
```

Each provider credential doesn't yet need to differentiate cloud/infra so they just straight inherit from the template credential:

```
class ManageIQ::Providers::EmbeddedTerraform::AutomationManager::AmazonCredential < ManageIQ::Providers::EmbeddedTerraform::AutomationManager::TemplateCredential
class ManageIQ::Providers::EmbeddedTerraform::AutomationManager::VsphereCredential < ManageIQ::Providers::EmbeddedTerraform::AutomationManager::TemplateCredential
```

jrafanie: this was done in https://github.com/ManageIQ/manageiq-providers-embedded_terraform/pull/20

app/models/manageiq/providers/embedded_terraform/automation_manager/amazon_credential.rb
jrafanie: To be resolved in in https://github.com/ManageIQ/manageiq-providers-embedded_terraform/pull/21

app/models/manageiq/providers/embedded_terraform/automation_manager/ibm_cloud_credential.rb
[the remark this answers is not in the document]
jrafanie: To be resolved in https://github.com/ManageIQ/manageiq-providers-embedded_terraform/pull/21

PR #21, opened 2024-05-01
Simplify descriptions from credentials PR #16

jrafanie:
This PR simplifies a few descriptions from #16 

Updated screens:

image

Note, for the IBM cloud, it looks like we need either the cloud API key or both the classic infra user name and classic infra API key so it doesn't make sense to require the cloud API key.  Is this correct @putmanoj?

image


app/models/manageiq/providers/embedded_terraform/automation_manager/ibm_cloud_credential.rb
```diff
@@ -3,28 +3,26 @@ class ManageIQ::Providers::EmbeddedTerraform::AutomationManager::IbmCloudCredent
...
+      :helperText => N_('The User Name for IBM Cloud Classic Infrastructure. This credential requires either an IBM Cloud API Key or both a Classic Infrastructure User Name and Classic Infrastructure API key.'),
       :name       => 'classic_user',
       :id         => 'classic_user',
       :maxLength  => 100,
     },
     {
       :component  => 'password-field',
       :label      => N_('IBM Cloud Classic Infrastructure API Key'),
-      :helperText => N_('The API key for IBM Cloud Classic Infrastructure A value for this field is required when using classic IBM Cloud resources. A valid connection must have value for IBM Cloud API Key or, IBM Cloud Classic Infrastructure User Name and IBM Cloud Classic Infrastructure API Key.'),
+      :helperText => N_('The API key for IBM Cloud Classic Infrastructure. This credential requires either an IBM Cloud API Key or both a Classic Infrastructure User Name and Classic Infrastructure API key.'),
```
jrafanie, line 25: I don't like how these "what's required" are repeated.  @GilbertCherrie do we have a way to say we require either:

auth_key OR
(classic_user AND classic_key)

I think the `:validate => [{:type => 'required'}]` I removed above is to simplistic for this logic.

app/models/manageiq/providers/embedded_terraform/automation_manager/ibm_cloud_credential.rb
```diff
@@ -3,28 +3,26 @@ class ManageIQ::Providers::EmbeddedTerraform::AutomationManager::IbmCloudCredent
     {
       :component  => 'password-field',
       :label      => N_('IBM Cloud API Key'),
-      :helperText => N_('The API key for IBM Cloud. A value for this field is required if classic user name and classic API key are not provided. A valid connection must have value for IBM Cloud API Key or, IBM Cloud Classic Infrastructure User Name and IBM Cloud Classic Infrastructure API Key.'),
+      :helperText => N_('The API key for IBM Cloud. This credential requires either an IBM Cloud API Key or both a Classic Infrastructure User Name and Classic Infrastructure API key.'),
```
Fryguy, line 6: Does something like this work?

```suggestion
      :helperText => N_('The API key for IBM Cloud. This credential requires either an IBM Cloud API Key or a Classic Infrastructure User Name / API key pair.'),
```

Fryguy: Or perhaps

Does something like this work?

```suggestion
      :helperText => N_('The API key for IBM Cloud. This credential requires either an IBM Cloud API Key or a Classic Infrastructure User Name and API key.'),
```

jrafanie: yeah, English vs. logical makes that not obvious.  I think I like the second one.  I'm not a fan of duplicating the same thing 3 times here so do wish we could put this logic into the validation for this form though so we can help the user if they don't provide the right things.

Fryguy: Agreed - maybe some tabs?  like we do for multiple auths?

Fryguy:
actually, separate credential types feels easier

app/models/manageiq/providers/embedded_terraform/automation_manager/ibm_cloud_credential.rb
Fryguy: actually, separate credential types feels easier

jrafanie: I agree, let me split off this into a new PR and get the other change in.  I'll create a new credential type and play with that.

jrafanie:
**NOTE: I'll split off the IBM Cloud credentials into a followup PR with two types: cloud and classic infrastructure.**

Note, for the IBM cloud, it looks like we need either the cloud API key or both the classic infra user name and classic infra API key so it doesn't make sense to require the cloud API key.  Is this correct @putmanoj?

image

app/models/manageiq/providers/embedded_terraform/automation_manager/ibm_cloud_credential.rb
jrafanie: Moving this to a followup PR with 2 credential types.

app/models/manageiq/providers/embedded_terraform/automation_manager/ibm_cloud_credential.rb
jrafanie: Ok, split it off into a separate class.  This will allow us to test both IBM Cloud and IBM Classic Infrastructure: https://github.com/ManageIQ/manageiq-providers-embedded_terraform/pull/23

PR #23, opened 2024-05-03
Split ibm cloud credentials from ibm classic infra credentials

jrafanie:
I think this simplifies the required field logic.  We'll need to make sure connecting from within the runner works but this is the start of enabling that work.

image

image

image

image

### DIFF
--- a/app/models/manageiq/providers/embedded_terraform/automation_manager/ibm_classic_infrastructure_credential.rb
+++ b/app/models/manageiq/providers/embedded_terraform/automation_manager/ibm_classic_infrastructure_credential.rb
@@ -28,11 +28,11 @@ class ManageIQ::Providers::EmbeddedTerraform::AutomationManager::IbmClassicInfra
 
   API_OPTIONS = {
     :type       => 'cloud',
-    :label      => N_('IBM Classic Infrastructure'),
+    :label      => N_('IBM Cloud Classic Infrastructure'),
     :attributes => API_ATTRIBUTES
   }.freeze
 
   def self.display_name(number = 1)
-    n_('Credential (IBM Classic Infrastructure)', 'Credentials (IBM Classic Infrastructure)', number)
+    n_('Credential (IBM Cloud Classic Infrastructure)', 'Credentials (IBM Cloud Classic Infrastructure)', number)
   end
 end
--- a/app/models/manageiq/providers/embedded_terraform/automation_manager/ibm_classic_infrastructure_credential.rb
+++ b/app/models/manageiq/providers/embedded_terraform/automation_manager/ibm_classic_infrastructure_credential.rb
@@ -1,0 +1,38 @@
+class ManageIQ::Providers::EmbeddedTerraform::AutomationManager::IbmClassicInfrastructureCredential < ManageIQ::Providers::EmbeddedTerraform::AutomationManager::TemplateCredential
+  COMMON_ATTRIBUTES = [].freeze
+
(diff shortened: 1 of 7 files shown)


app/models/manageiq/providers/embedded_terraform/automation_manager/ibm_cloud_credential.rb
```diff
@@ -42,17 +25,4 @@ class ManageIQ::Providers::EmbeddedTerraform::AutomationManager::IbmCloudCredent
...
-
-  def self.params_to_attributes(params)
-    attrs = super.dup
-
-    attrs[:auth_key] = attrs.delete(:auth_key) if attrs.key?(:auth_key)
-
-    if %i[classic_user classic_key].any? { |opt| attrs.key?(opt) }
-      attrs[:userid]   = attrs.delete(:classic_user) if attrs.key?(:classic_user)
-      attrs[:password] = attrs.delete(:classic_key)  if attrs.key?(:classic_key)
-    end
```
jrafanie, line 54: I think this just goes away if classic user/key are moved to the other class as auth_key is required, right?

app/models/manageiq/providers/embedded_terraform/automation_manager/ibm_classic_infrastructure_credential.rb
```diff
@@ -0,0 +1,38 @@
...
+      :component  => 'password-field',
+      :label      => N_('IBM Cloud Classic Infrastructure API Key'),
+      :helperText => N_('The API key for IBM Cloud Classic Infrastructure.'),
+      :name       => 'password',
+      :id         => 'password',
+      :type       => 'password',
+      :isRequired => true,
+      :validate   => [{:type => 'required'}],
+    },
+  ].freeze
```
jrafanie, line 25: I renamed these to userid and password so we don't need to define `params_to_attributes` here.

agrare:
Makes sense to me, @putmanoj can you confirm this will work for opentofu-runner / ibm_cloud provider

app/models/manageiq/providers/embedded_terraform/automation_manager/ibm_cloud_credential.rb
Fryguy: Not sure - @agrare ?

Fryguy:
Only comment I have here is I think it should it be "IBM Cloud Classic Infrastructure" vs "IBM Classic Infrastructure"?  (it's this way in various strings, class names, factory names etc, so I think it would have to change everywhere)

jrafanie:
> Only comment I have here is I think it should it be "IBM Cloud Classic Infrastructure" vs "IBM Classic Infrastructure"? (it's this way in various strings, class names, factory names etc, so I think it would have to change everywhere)

My goal was to avoid confusing people by having IBM Cloud and IBM Cloud Classic Infrastructure in the select list.  I saw Classic Infrastructure all over the place here and figured maybe we could use it: https://cloud.ibm.com/docs/account?topic=account-classic_keys

jrafanie:
> Only comment I have here is I think it should it be "IBM Cloud Classic Infrastructure" vs "IBM Classic Infrastructure"? (it's this way in various strings, class names, factory names etc, so I think it would have to change everywhere)

Ok, renamed to IBM Cloud Classic Infrastructure.  The screenshots are updated.

jrafanie:
Ready to go

PR #24, opened 2024-05-13
Attaching volume to OpenTofu runner to read https certs

sudhir-kelkar:
This PR will mount cert path to opentofu-runner worker deployment by reading certs from secret `application-certificate-secret`

### DIFF
--- a/app/models/opentofu_worker.rb
+++ b/app/models/opentofu_worker.rb
@@ -49,6 +49,13 @@ class OpentofuWorker < MiqWorker
     }
   end
 
+  def configure_service_worker_deployment(definition)
+    super
+
+    definition[:spec][:template][:spec][:containers].first[:volumeMounts] << {:name => "cert-path", :mountPath => "/root"}
+    definition[:spec][:template][:spec][:volumes] << {:name => "cert-path", :secret => {:secretName => "application-certificate-secret", :items => [{:key => "httpd_crt", :path => "server.crt"}, {:key => "httpd_key", :path => "server.key"}], :defaultMode => 420}}
+  end
+
   def create_podman_secret
     return if AwesomeSpawn.run("runuser", :params => [[:login, "manageiq"], [:command, "podman secret exists --root=#{Rails.root.join("data/containers/storage")} opentofu-runner-secret"]]).success?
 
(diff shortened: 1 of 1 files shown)


Fryguy:
@bdunne Please review.

app/models/opentofu_worker.rb
```diff
@@ -49,6 +49,13 @@ def unit_environment_variables
     }
   end
 
+  def configure_service_worker_deployment(definition)
+    super
+
+    definition[:spec][:template][:spec][:containers].first[:volumeMounts] << {:name => "cert-path", :mountPath => "/root"}
```
bdunne, line 55: Does the container have access to `/root`?

app/models/opentofu_worker.rb
```diff
@@ -49,6 +49,13 @@ def unit_environment_variables
...
+    definition[:spec][:template][:spec][:volumes] << {:name => "cert-path", :secret => {:secretName => "application-certificate-secret", :items => [{:key => "httpd_crt", :path => "server.crt"}, {:key => "httpd_key", :path => "server.key"}], :defaultMode => 420}}
```
bdunne, line 56: `application-certificate-secret` should not be hard-coded.

bdunne: `{:key => "httpd_crt"` This should have its own certificate, not reuse the one for httpd.

app/models/opentofu_worker.rb
sudhir-kelkar: Using this path now which is accessible from terraform runner container `/opt/app-root/src/`
Writing certificate and key on path `/opt/app-root/src/config/cert`

app/models/opentofu_worker.rb
sudhir-kelkar: Updated IM installer code to have new key and cert in secret `application-certificate-secret`.
Using those secret key for creating new volume mounts.
ref: https://github.ibm.com/IBMPrivateCloud/ibm-infra-management-install-operator/pull/566/files

sudhir-kelkar: There is no environment variable I can use for reading secret name as `application-certificate-secret`. 
May be I am missing existing so please guide. Secondly If I define the variable and assign value to it as `application-certificate-secret`  it is as good as hard coding.

sudhir-kelkar: Using environment variable API_SSL_SECRET_NAME instead of actual secret name

PR #27, opened 2024-05-15
Fix OpentofuWorker worker_deployment_name

agrare:
The systemd service name is "opentofu-runner" so it would make sense to have the podified deployment match.

### DIFF
--- a/app/models/opentofu_worker.rb
+++ b/app/models/opentofu_worker.rb
@@ -13,12 +13,12 @@ class OpentofuWorker < MiqWorker
     "#{service_base_name}.service"
   end
 
-  def self.worker_deployment_name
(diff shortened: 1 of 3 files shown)


agrare:
Unrelated but the `MiqWorker::ContainerCommon#worker_deployment_name` instance method goes straight from `abbreviated_class_name` to the `region-underscoreized-ems_id`.  If we had a class-level worker_deployment_base_name that would make it easier to override without having to copy the region+ems_id bits.

bdunne (commented):
Can you add a test for the name (and anything else you think is important) in the deployment definition?

bdunne (approved):
:+1: Awesome! Thanks so much for the tests

PR #29, opened 2024-05-22
Override default containerport 3000 to opentofu runner port 6000

sudhir-kelkar:
Container port defined in worker core deployment teamplate is 3000.
But we want opentofu-runner to work on 6000 port. Overriding the same in embedded plugin

### DIFF
--- a/app/models/opentofu_worker.rb
+++ b/app/models/opentofu_worker.rb
@@ -5,8 +5,6 @@ class OpentofuWorker < MiqWorker
   self.rails_worker          = false
   self.maximum_workers_count = 1
 
-  STARTING_PORT = 6000
-
   def self.service_base_name
     "opentofu-runner"
   end
@@ -24,7 +22,7 @@ class OpentofuWorker < MiqWorker
   end
 
   def container_port
-    6001
+    6000
   end
 
   private
--- a/app/models/opentofu_worker.rb
+++ b/app/models/opentofu_worker.rb
@@ -5,6 +5,8 @@ class OpentofuWorker < MiqWorker
   self.rails_worker          = false
   self.maximum_workers_count = 1
 
+  STARTING_PORT = 6000
+
   def self.service_base_name
     "opentofu-runner"
   end
@@ -19,6 +21,10 @@ class OpentofuWorker < MiqWorker
 
   def worker_deployment_name
     @worker_deployment_name ||= "#{deployment_prefix}opentofu-runner"
+  end
+
+  def container_port
+    6001
   end
 
   private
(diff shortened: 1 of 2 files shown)


app/models/opentofu_worker.rb
```diff
@@ -21,6 +23,10 @@ def worker_deployment_name
     @worker_deployment_name ||= "#{deployment_prefix}opentofu-runner"
   end
 
+  def container_port
+    6001
```
bdunne, line 27: Shouldn't this be 6000 to match the service? https://github.com/ManageIQ/manageiq-pods/pull/1108/files#diff-3e26d51b3b228e4cc4a9256f6b4a49c75730fb226bb062aa45ff405040313608R29

app/models/opentofu_worker.rb
```diff
@@ -5,6 +5,8 @@ class OpentofuWorker < MiqWorker
   self.rails_worker          = false
   self.maximum_workers_count = 1
 
+  STARTING_PORT = 6000
```
bdunne, line 8: I think this is only used by our rails workers, right?

app/models/opentofu_worker.rb
sudhir-kelkar: This is how it is in MiqUiWorker. Need to see why? but with this changes container starting on 6000

Fryguy: As discussed in chat, this is 3001 for UI worker because that pod also has an internal httpd process that runs on 3000, so we need this method to tells the rails runner what port to start on, since it's not 3000.

I'm not sure if this would be used at all by the opentofu runner unless there's code to tell the opentofu-runner container to run on that port.

app/models/opentofu_worker.rb
agrare: This is only used by `MiqWebServerWorkerMixin` so yeah we probably don't need this

app/models/opentofu_worker.rb
sudhir-kelkar: This container port sets environment variable `PORT` on the container. Opentofu runner express server reads from this env var.
I am in process of testing my changes on the server.. will check this behavior to confirm.

Fryguy:
This just missed the merge of master into radjabov so marking for backport

Fryguy:
Backported to `radjabov` in commit 49993907dc9f42e0b3977a4f3c15e895d80b9f8d.

```text
commit 49993907dc9f42e0b3977a4f3c15e895d80b9f8d
Author: Brandon Dunne <brandondunne@hotmail.com>
Date:   Wed May 22 17:58:04 2024 -0400

    Merge pull request #29 from sudhir-kelkar/createOpentofuContainerPort
    
    Override default containerport 3000 to opentofu runner port 6000
    
    (cherry picked from commit f8e30f095811d4f1d5827ebf9f785fd0a0a83f04)
```

PR #47, opened 2024-05-29
[WIP] Raise an error if terraform runner run fails

### LINKED ISSUE #45: No error message if Terraform::Runner.run fails
`Terraform::Runner.run` takes a `Terraform::Runner::Response` and returns `Terraform::Runner::ResponseAsync.new(response.stack_id)`

If the Response was a failure then `stack_id` will be `nil`, and Job.options will have a `nil` terraform_stack_id
` options: {:template_id=>2, :env_vars=>{}, :input_vars=>{"execution_ttl"=>"", "extra_vars"=>{}, "verbosity"=>"0"}, :credentials=>[2], :poll_interval=>1 minute, :git_checkout_tempdir=>"/tmp/embedded-terraform-runner-git20240528-18757-aa2dl6", :terraform_stack_id=>nil}`

agrare:
If the Terraform::Runner.run call fails there is no stack_id returned which means that any error message returned by the opentofu-runner API is lost.

Re-work how the API response is handled so that the error message is available if the initial call fails.

Fixes https://github.com/ManageIQ/manageiq-providers-embedded_terraform/issues/45

### DIFF
--- a/lib/terraform/runner.rb
+++ b/lib/terraform/runner.rb
@@ -5,6 +5,9 @@ require 'base64'
 
 module Terraform
   class Runner
+    class Error < StandardError
+    end
+
     class << self
       def available?
         return @available if defined?(@available)
@@ -25,8 +28,8 @@ module Terraform
       # @param env_vars [Hash] Hash with key/value pairs that will be passed as environment variables to the
       #        terraform-runner run
       # @return [Terraform::Runner::ResponseAsync] Response object of terraform-runner create action
-      def run(input_vars, template_path, tags: nil, credentials: [], env_vars: {})
-        _log.debug("Run_aysnc template: #{template_path}")
+      # @raise [Terraform::Runner::Error] Raises an exception if runnign the template fails
+      def run!(input_vars, template_path, tags: nil, credentials: [], env_vars: {})
         response = create_stack_job(
           template_path,
           :input_vars  => input_vars,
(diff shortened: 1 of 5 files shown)


lib/terraform/runner.rb
```diff
@@ -25,28 +28,44 @@ def available?
...
+      def run!(input_vars, template_path, tags: nil, credentials: [], env_vars: {})
         response = create_stack_job(
           template_path,
           :input_vars  => input_vars,
           :tags        => tags,
           :credentials => credentials,
           :env_vars    => env_vars
         )
+
+        raise Terraform::Runner::Error, response.error_message if response.status == "FAILED"
```
agrare, line 41: @putmanoj can you help out here?  I'm seeing in ResponseAsync that status is a string like: `"SUCCESS", "FAILED", "CANCELLED"`, but if I submit a create_stack_job I see status is 401:
```
#<Terraform::Runner::Response:0x00007f94dcb031c8
 @action=nil,
 @created_at=nil,
 @details=nil,
 @error_message=nil,
 @message="You do not have the necessary credentials.",
 @stack_id=nil,
 @stack_job_end_time=nil,
 @stack_job_start_time=nil,
 @stack_name=nil,
 @status=401>
```

agrare: Also, when is message used vs error_message?

putmanoj: > Also, when is message used vs error_message?

The 'message' is the full terraform/opentofu command output (including error message)  
whereas 'error_message' ,  in case fails, cause of the failure or it only contains the error message for the failure.

agrare: Okay, I have a failure posted above and there is only an entry in `message` and not `error_message` is that normal?

putmanoj: I suppose we can have some improvement here.
But in this case the terraform command was not executed, as there auth failure when calling the API.

For now to complete the dev test, you could use this example 'TOKEN' -> https://github.ibm.com/Orpheus/terraform-runner/blob/main/examples/simple-template-hcl/01-create-stack.sh#L11

agrare: @putmanoj this isn't dev this is on an appliance, what is the plan for auth in prod?

putmanoj: For now you could hardcode the above token and continue.

We can continue the discussion on the channel

agrare: Opened to track this issue https://github.com/ManageIQ/manageiq-providers-embedded_terraform/issues/49

agrare: @putmanoj circling back to this question, I know we are tackling the auth issue separately but how do I tell when a terraform response is success/failure and how do I get the error response?

PR #51, opened 2024-05-31
Updating opentofu-runner worker deployment as per terraform-runner service requirement

sudhir-kelkar:
In this pull request, I am updating the opentofu runner worker deployment as per the Terraform runner service requirements:

- Adding runAsUser to match the node user from opentofu-runner. Setting it to `1000690001`.
- Setting multiple mount paths required by Terraform to write its execution files, such as `terraform-bin-empty`, `terraform-d-empty`, and `stacks-empty`.
- Updating the HOME variable from /var/www/miq/vmdb/tmp to /home/node.

### DIFF
--- a/app/models/opentofu_worker.rb
+++ b/app/models/opentofu_worker.rb
@@ -74,10 +74,24 @@ class OpentofuWorker < MiqWorker
 
   def configure_service_worker_deployment(definition)
     super
-
+    # overwriting container port to be same as opentofu-runner service port i.e. in this case 6000
     definition[:spec][:template][:spec][:containers].first[:ports] = [{:containerPort => container_port}]
 
+    # ovewriting home directory to terraform home dir
+    env_var_array = definition[:spec][:template][:spec][:containers][0][:env]
+    env_var_array.detect { |env| env[:name] == "HOME" }&.[]=(:value, "/home/hode")
+
+    definition[:spec][:template][:spec][:containers].first[:securityContext][:runAsUser] = 1_000_690_001
+    # these volume mounts are require by terraform runner to create the stack, mentioned it as {} so that it can be writable
+    definition[:spec][:template][:spec][:containers].first[:volumeMounts] << {:name => "terraform-bin-empty", :mountPath => "/home/node/terraform/bin"}
+    definition[:spec][:template][:spec][:volumes] << {:name => "stacks-empty", :emptyDir => {}}
+    definition[:spec][:template][:spec][:containers].first[:volumeMounts] << {:name => "terraform-d-empty", :mountPath => "/home/node/terraform/.terraform.d"}
+    definition[:spec][:template][:spec][:volumes] << {:name => "terraform-bin-empty", :emptyDir => {}}
+    definition[:spec][:template][:spec][:containers].first[:volumeMounts] << {:name => "stacks-empty", :mountPath => "/stacks"}
+    definition[:spec][:template][:spec][:volumes] << {:name => "terraform-d-empty", :emptyDir => {}}
+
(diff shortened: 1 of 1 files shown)


agrare:
@sudhir-kelkar please drop any "test" commits from the PR so they don't make it into the master git history, https://github.com/ManageIQ/manageiq-providers-embedded_terraform/pull/51/commits/475f18117634884d0d9c85847539465b676a570b

app/models/opentofu_worker.rb
```diff
@@ -74,10 +74,32 @@ def unit_environment_variables
...
+      if env_var_key_val[:name] == var_name
+        env_var_key_val[:value] = updated_value
+        break
+      end
+    end
+
+    # setting up runasuser
+    definition[:spec][:template][:spec][:containers][0][:securityContext][:runAsUser] = 1000690001
+    # these volume mounts are require by terraform runner to create the stack, mentioned it as {} so that it can be writable
+    definition[:spec][:template][:spec][:containers].first[:volumeMounts] << {:name => "terraform-bin-empty", :mountPath => "/home/node/terraform/bin"}
```
agrare: Just cosmetic but can we standardize on either `[:containers][0]` or `[:containers].first` here?

I see we use both in core, but in the same method it'd be nice if we used the same one.

app/models/opentofu_worker.rb
```diff
@@ -74,10 +74,32 @@ def unit_environment_variables
...
+    var_name = "HOME"
+    updated_value = "/home/node"
+    env_var_array.each do |env_var_key_val|
+      if env_var_key_val[:name] == var_name
+        env_var_key_val[:value] = updated_value
+        break
+      end
+    end
+
+    # setting up runasuser. specifying '_' to avoid lint issue
```
agrare, line 91: > `specifying '_' to avoid lint issue` 

I don't think we need this comment, this is a very common pattern in ManageIQ (and ruby generally)

app/models/opentofu_worker.rb
```diff
@@ -74,10 +74,32 @@ def unit_environment_variables
...
+    # ovewriting home directory to terraform home dir
+    env_var_array = definition[:spec][:template][:spec][:containers][0][:env]
+    var_name = "HOME"
+    updated_value = "/home/node"
+    env_var_array.each do |env_var_key_val|
+      if env_var_key_val[:name] == var_name
+        env_var_key_val[:value] = updated_value
+        break
+      end
+    end
```
agrare, line 89: I'm fine changing this here but I think a good follow-up would be to allow overriding of a home_dir method so that the default_environment used that rather than `Rails.root.join("tmp").to_s}`

agrare: You can use detect to find a hash value so you don't have to write your own loop/break logic
```suggestion
    env_var_array = definition[:spec][:template][:spec][:containers][0][:env]
    env_var_array.detect { |env| env[:name] == "HOME" }[:value] = "/home/node"
```

sudhir-kelkar: Thanks... I have updated the PR with your suggested code

agrare (approved):

Fryguy:
Backported to `radjabov` in commit f397226486c9ba1432e3c6162890d922cef1b183.

```text
commit f397226486c9ba1432e3c6162890d922cef1b183
Date:   Sat Jun 1 09:21:12 2024 -0400

    Merge pull request #51 from sudhir-kelkar/createSecretKeyEnv
    
    Updating opentofu-runner worker deployment as per terraform-runner service requirement
    
    (cherry picked from commit 305bbae34ce31d0e1c1027cd1082d4aae023b115)
```

Issue #38, opened 2024-05-24
Opentofu Runner Systemd unit doesn't shutdown cleanly


agrare:
```
podman[14078]: [2024-05-24T16:17:00.436] [INFO] wait-for-services - All required services now available
systemd[1]: Stopping opentofu-runner.service...
podman[14582]: time="2024-05-24T12:40:20-04:00" level=warning msg="StopSignal SIGTERM failed to stop container opentofu-runner in 30 seconds, resorting to SIGKILL"
podman[14078]: 2024-05-24 12:40:20.777734677 -0400 EDT m=+1402.420094741 container died 2f7a904bfecf42d0e2f88a1c86d493a50e396093e4709c41d8c10fc8a8afd153 (image=docker-na-public.artifactory.swg-devops.com/hyc-cp4mcm-team-docker-local/infra-mgmt/infra-mgmt-opentofu-runner:PR-110-204033d-20240521-x86_64, name=opentofu-runner, io.openshift.expose-services=8080:http, io.s2i.scripts-url=image:///usr/libexec/s2i, url=https://access.redhat.com/containers/#/registry.access.redhat.com/ubi9/nodejs-18-minimal/images/1-113.1715236126, help=For more information visit https://github.com/sclorg/s2i-nodejs-container, vcs-type=git, release=113.1715236126, io.buildah.version=1.29.0, io.k8s.display-name=Opentofu runner, vcs-ref=90fac532ef0e8a98b1962a72369005dcf8c321ea, com.redhat.component=nodejs-18-minimal-container, vendor=IBM, com.redhat.deployments-dir=/opt/app-root/src, build-date=2024-05-09T06:29:39, com.redhat.license_terms=https://www.redhat.com/en/about/red-hat-end-user-license-agreements#UBI, maintainer=SoftwareCollections.org <sclorg@redhat.com>, io.openshift.tags=Opentofu runner, name=Opentofu runner, architecture=x86_64, distribution-scope=public, summary=Opentofu runner, version=1, com.redhat.dev-mode.port=DEBUG_PORT:5858, io.openshift.s2i.scripts-url=image:///usr/libexec/s2i, com.redhat.dev-mode=DEV_MODE:false, io.k8s.description=Opentofu runner, description=Opentofu runner)
[...]

PR #52, opened 2024-05-31
Fix some issues with opentofu-runner systemd service

agrare:
Fix issues with --cgroup=systemd missing systemd user session
```
The cgroupv2 manager is set to systemd but there is no systemd user session available
For using systemd, you may need to log in using a user session
Alternatively, you can enable lingering with: `loginctl enable-linger 985` (possibly as root)
Falling back to --cgroup-manager=cgroupfs
```

Also move our service file closer to what would be generated by the `Quadlet` systemd generator in https://github.com/ManageIQ/manageiq-providers-embedded_terraform/pull/50

We aren't able to use that entirely yet but this ~~is inspired by~~ blatantly copies the resulting transient service file.

I split out the fix for each issue into a separate commit for easier review

### DIFF
--- a/systemd/opentofu-runner.service
+++ b/systemd/opentofu-runner.service
@@ -5,7 +5,7 @@ WantedBy=opentofu-runner.target
 [Service]
 User=manageiq
 Group=manageiq
-ExecStartPre=/bin/rm -f /tmp/%n.cid
+ExecStartPre=/bin/rm -f %T/%N.pid %T/%N.cid
 ExecStartPre=loginctl enable-linger manageiq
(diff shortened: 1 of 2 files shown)


systemd/opentofu-runner.service
```diff
@@ -1,16 +1,25 @@
...
+Environment=PODMAN_SYSTEMD_UNIT=%n
+KillMode=mixed
+Delegate=yes
+Type=notify
+NotifyAccess=all
+SyslogIdentifier=%N
+ExecStartPre=/bin/rm -f %T/%N.cid
+ExecStartPre=loginctl enable-linger manageiq
+ExecStart=/usr/bin/podman run --name=opentofu-runner --cidfile=%T/%N.cid --replace --rm --cgroups=split --network=host --sdnotify=conmon -d -v /var/lib/manageiq/opentofu-runner/certs:/opt/app-root/src/config/cert:z --env DATABASE_HOSTNAME=${DATABASE_HOSTNAME} --env DATABASE_NAME=${DATABASE_NAME} --env DATABASE_USERNAME=${DATABASE_USERNAME} --env MEMCACHE_SERVERS=${MEMCACHE_SERVERS} --env PORT=${PORT} --expose ${PORT} --secret opentofu-runner-secret --root /var/www/miq/vmdb/data/containers/storage ${OPENTOFU_RUNNER_IMAGE}
+ExecStop=/usr/bin/podman kill --root /var/www/miq/vmdb/data/containers/storage --cidfile=%T/%N.cid
```
agrare, line 24: NOTE I switched to podman kill here because the opentofu-runner image doesn't stop cleanly causing timeout and a SIGKILL anyway, we should fix this and is being tracked by https://github.com/ManageIQ/manageiq-providers-embedded_terraform/issues/38

Fryguy:
Backported to `radjabov` in commit 3ffb2bdc0ae4ced0b4654c866ec618b2fa9e2421.

```text
commit 3ffb2bdc0ae4ced0b4654c866ec618b2fa9e2421
Date:   Mon Jun 3 14:01:59 2024 -0400

    Merge pull request #52 from agrare/improve_opentofu_worker_systemd_service
    
    Fix some issues with opentofu-runner systemd service
    
    (cherry picked from commit d07d56520b76ebf6daf6f1f6b39d1982e1092f3c)
```

PR #53, opened 2024-06-03
Parse dialog options as Extra Vars

agrare:
When you have a dialog pass a value for a service order it comes in as:
```
options
{:dialog=>{"dialog_text_box_1"=>"hello world"}
```

We need to extract this and pass through to the Terraform::Runner as Extra Vars

With this change the Job options now look like:
```
 options:
  {:template_id=>4,
   :env_vars=>{},
   :input_vars=>{"execution_ttl"=>"", "extra_vars"=>{"text_box_1"=>"hello world"}, "verbosity"=>"0"},
   :credentials=>[3],
   :poll_interval=>1 minute,
   :git_checkout_tempdir=>"/tmp/embedded-terraform-runner-git20240603-334161-40j3os"},
```

### DIFF
--- a/app/models/manageiq/providers/embedded_terraform/automation_manager/job.rb
+++ b/app/models/manageiq/providers/embedded_terraform/automation_manager/job.rb
@@ -21,13 +21,9 @@ class ManageIQ::Providers::EmbeddedTerraform::AutomationManager::Job < Job
   def execute
     template_path = File.join(options[:git_checkout_tempdir], template_relative_path)
     credentials   = Authentication.where(:id => options[:credentials])
+    extra_vars    = options.dig(:input_vars, :extra_vars) || {}
 
-    response = Terraform::Runner.run(
-      options[:input_vars],
-      template_path,
-      :credentials => credentials,
-      :env_vars    => options[:env_vars]
-    )
+    response = Terraform::Runner.run(decrypt_extra_vars(extra_vars), template_path, :credentials => credentials, :env_vars => options[:env_vars])
 
     options[:terraform_stack_id] = response.stack_id
     save!
@@ -108,6 +104,11 @@ class ManageIQ::Providers::EmbeddedTerraform::AutomationManager::Job < Job
     return if options[:terraform_stack_id].nil?
 
     @stack_response ||= Terraform::Runner::ResponseAsync.new(options[:terraform_stack_id])
+  end
+
+  def decrypt_extra_vars(extra_vars)
+    result = extra_vars.deep_dup
+    result.transform_values! { |val| val.kind_of?(String) ? ManageIQ::Password.try_decrypt(val) : val }
   end
 
   def configuration_script_source
--- a/app/models/service_terraform_template.rb
+++ b/app/models/service_terraform_template.rb
@@ -85,7 +85,7 @@ class ServiceTerraformTemplate < ServiceGeneric
     job_options[:extra_vars].try(:transform_values!) do |val|
       val.kind_of?(String) ? val : val[:default] # TODO: support Hash only
     end
-    # TODO: job_options.deep_merge!(parse_dialog_options) unless action == ResourceAction::RETIREMENT
+    job_options.deep_merge!(parse_dialog_options) unless action == ResourceAction::RETIREMENT
     job_options.deep_merge!(overrides)
     translate_credentials!(job_options)
 
@@ -95,6 +95,17 @@ class ServiceTerraformTemplate < ServiceGeneric
 
   def job_option_key(action)
     "#{action.downcase}_job_options".to_sym
+  end
+
+  def parse_dialog_options
+    dialog_options = options[:dialog] || {}
+
+    params = dialog_options.each_with_object({}) do |(attr, val), obj|
+      var_key = attr.sub(/^(password::)?dialog_/, '')
+      obj[var_key] = val
+    end
+
+    params.blank? ? {} : {:extra_vars => params}
   end
 
   def translate_credentials!(options)


app/models/service_terraform_template.rb
```diff
@@ -97,6 +97,17 @@ def job_option_key(action)
     "#{action.downcase}_job_options".to_sym
   end
 
+  def parse_dialog_options
+    dialog_options = options[:dialog] || {}
+
+    params = dialog_options.each_with_object({}) do |(attr, val), obj|
+      var_key = attr.sub(/^(password::)?dialog_/, '')
```
Fryguy, line 104: are the password values encrypted?  If so, do they need to be decrypted?

app/models/service_terraform_template.rb
```diff
@@ -97,6 +97,17 @@ def job_option_key(action)
...
+      obj[var_key] = val unless var_key == attr
```
Fryguy, line 105: Is this saying to _not_ clobber the value if it's already an extra var?  If so, I'm not sure that's right.  Typically the dialog overrides any hardcoded values.

agrare: So background most of this method is pulled form embedded ansible, https://github.com/ManageIQ/manageiq/blob/master/app/models/service_ansible_playbook.rb#L167

I don't think that is the intent, since this is only looking at options[:dialog] and that isn't where hard-coded extra vars are stored.

I'm trying to dig into why ansible does this and the password decryption, we might be able to drop both and just go with a simpler approach.

agrare: Ah okay I thought that was an ansible specific thing but protected dialog fields look like `"password::dialog_password"=>"v2:{1aCicMVgRQFLeGN5K/LWcA==}"}`

I'm not sure that I like where the decrypt is done with embedded_ansible because it looks like the Job.options will have the plaintext value...

Fryguy: right, the password:: prefix is there to ensure that we don't print it in logs (if we see password in the key we are certain it's a sensitive value)

agrare: This was added to embedded_ansible here https://github.com/ManageIQ/manageiq/pull/14636 and it is decrypted here https://github.com/ManageIQ/manageiq/pull/14636/files#diff-43965f59b3501c850937c667f9712b5d46a04da94da5385d74f2fc187f170ca0R16

I've opted to decrypt right before sending to Terraform::Runner so the plaintext values don't make it to the Job.options column.  I might be missing something on the embedded_ansible side maybe they're re-encrypting them before creating the Job but this seemed like a good way to go.

agrare: That was also where the check for `unless var_key == attr` maybe so that password and non-password text boxes don't overwrite each other?

app/models/service_terraform_template.rb
agrare: They can be, and yes they do need to be

app/models/manageiq/providers/embedded_terraform/automation_manager/job.rb
```diff
@@ -110,6 +106,13 @@ def stack_response
     @stack_response ||= Terraform::Runner::ResponseAsync.new(options[:terraform_stack_id])
   end
 
+  def decrypt_input_vars(input_vars)
+    input_vars
+      .dup
+      .fetch("extra_vars", {})
+      .transform_values { |val| val.kind_of?(String) ? ManageIQ::Password.try_decrypt(val) : val }
```
Fryguy, line 113: This dup does not work as expected (it only dups the outer hash, but not the inner hash), however it's also not needed since transform_values will create a new hash anyway:

```
[1] pry(main)> input_vars = {"extra_vars" => {}}
[2] pry(main)> [input_vars.object_id, input_vars["extra_vars"].object_id]
=> [320, 340]
[3] pry(main)> iv2 = input_vars.dup
[4] pry(main)> [iv2.object_id, iv2["extra_vars"].object_id]
=> [360, 340]
[5] pry(main)> ev = iv2.fetch("extra_vars", {})
[6] pry(main)> ev.object_id
=> 340
[7] pry(main)> ev2 = ev.transform_values {}
[8] pry(main)> ev2.object_id
=> 400
```

agrare:
```
?>   def decrypt_input_vars(input_vars)
?>     result = input_vars.deep_dup
?>     result.fetch("extra_vars", {}).transform_values! { |val| val.kind_of?(String) ? ManageIQ::Password.try_decrypt(val) : val }
?>     result
>>   end
=> :decrypt_input_vars
>> input_vars = {"extra_vars" => {"foo" => "bar", "bar" => ManageIQ::Password.encrypt("baz")}}
=> {"extra_vars"=>{"foo"=>"bar", "bar"=>"v2:{8v956e2OHCGXgGW7Q7xxDw==}"}}
>> decrypt_input_vars(input_vars)
=> {"extra_vars"=>{"foo"=>"bar", "bar"=>"baz"}}
>> input_vars
=> {"extra_vars"=>{"foo"=>"bar", "bar"=>"v2:{8v956e2OHCGXgGW7Q7xxDw==}"}}
```

I want to avoid modifying the original options and saving a decrypted value, went with deep_dup

Fryguy:
@agrare This now returns a different thing than the original commit (can't tell if intentional or not)

```ruby
  def decrypt_input_vars(input_vars)
    result = input_vars.deep_dup
    result.fetch("extra_vars", {}).transform_values! { |val| val.kind_of?(String) ? ManageIQ::Password.try_decrypt(val) : val }
    result
  end

  def decrypt_input_vars_orig(input_vars)
    input_vars
      .dup
      .fetch("extra_vars", {})
      .transform_values { |val| val.kind_of?(String) ? ManageIQ::Password.try_decrypt(val) : val }
  end
```

```
[7] pry(main)> decrypt_input_vars("extra_vars" => {"foo" => "bar"}, "flarp" => {})
=> {"extra_vars"=>{"foo"=>"bar"}, "flarp"=>{}}
[8] pry(main)> decrypt_input_vars_orig("extra_vars" => {"foo" => "bar"}, "flarp" => {})
=> {"foo"=>"bar"}
```

agrare:
@Fryguy I originally intended to return the entire input_vars payload

app/models/manageiq/providers/embedded_terraform/automation_manager/job.rb
```diff
@@ -21,13 +21,9 @@ def pre_execute
   def execute
     template_path = File.join(options[:git_checkout_tempdir], template_relative_path)
     credentials   = Authentication.where(:id => options[:credentials])
+    input_vars    = decrypt_input_vars(options[:input_vars])
```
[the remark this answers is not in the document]
agrare: Not seeing anywhere that logs the input_vars, https://github.com/ManageIQ/manageiq-providers-embedded_terraform/blob/master/lib/terraform/runner.rb#L29

Here is what it looks like in the logs with debug on
```
[----] I, [2024-06-03T15:07:06.459082#457302:9ee8]  INFO -- evm: MIQ(MiqQueue#deliver) Message id: [240], Delivering...
[----] I, [2024-06-03T15:07:06.494993#457302:9ee8]  INFO -- evm: MIQ(ManageIQ::Providers::EmbeddedTerraform::AutomationManager::Job#checkout_git_repository) Checking out git repository to "/tmp/embedded-terraform-runner-git20240603-457302-yitu28"...
[----] I, [2024-06-03T15:07:06.610938#457302:9ee8]  INFO -- evm: MIQ(GitRepository#update_repo) Updating git@github.ibm.com:manoj-puthran/cam-sample-templates.git in /home/grare/adam/src/manageiq/manageiq/data/git_repos/1...
[----] I, [2024-06-03T15:07:06.971855#457302:9ee8]  INFO -- evm: MIQ(GitRepository#update_repo) Updating git@github.ibm.com:manoj-puthran/cam-sample-templates.git in /home/grare/adam/src/manageiq/manageiq/data/git_repos/1......Complete
[----] I, [2024-06-03T15:07:06.972215#457302:9ee8]  INFO -- evm: MIQ(GitRepository#checkout) Checking out git@github.ibm.com:manoj-puthran/cam-sample-templates.git@master to /tmp/embedded-terraform-runner-git20240603-457302-yitu28...
[----] I, [2024-06-03T15:07:06.981529#457302:9ee8]  INFO -- evm: MIQ(GitRepository#checkout) Checking out git@github.ibm.com:manoj-puthran/cam-sample-templates.git@master to /tmp/embedded-terraform-runner-git20240603-457302-yitu28......Complete
[----] D, [2024-06-03T15:07:07.059332#457302:9ee8] DEBUG -- evm: MIQ(Terraform::Runner.run_async) Run_aysnc template: /tmp/embedded-terraform-runner-git20240603-457302-yitu28/terraform/hello-world
[----] I, [2024-06-03T15:07:07.059475#457302:9ee8]  INFO -- evm: MIQ(Terraform::Runner.create_stack_job) start stack_job for template: /tmp/embedded-terraform-runner-git20240603-457302-yitu28/terraform/hello-world
[----] D, [2024-06-03T15:07:07.059719#457302:9ee8] DEBUG -- evm: MIQ(Terraform::Runner.encoded_zip_from_directory) Create #<File:0x00007ff4f3e81ff8>
[----] D, [2024-06-03T15:07:07.059865#457302:9ee8] DEBUG -- evm: MIQ(Terraform::Runner.encoded_zip_from_directory) Adding camtemplate.json
[----] D, [2024-06-03T15:07:07.059969#457302:9ee8] DEBUG -- evm: MIQ(Terraform::Runner.encoded_zip_from_directory) Adding camvariables.json
[----] D, [2024-06-03T15:07:07.060044#457302:9ee8] DEBUG -- evm: MIQ(Terraform::Runner.encoded_zip_from_directory) Adding main.tf
```

No input_vars printed (thought the `#<File:0x00007ff4f3e81ff8>` looks like a bug)

> I think Ansible::Runner decrypts inside of itself as well, right before it sends it over, but I could be wrong

I don't see anything in Ansible::Runner calling ManageIQ::Password

agrare:
Actually looking at the Terraform::Runner.run_async docs they might want the extra_vars passed in rather than the whole input_vars

agrare:
Okay updated, should be good to go now.

Fryguy (approved):

Fryguy:
Backported to `radjabov` in commit 4830ad3a3f2ccfd8cc541ed43a04fdb158ee9cab.

```text
commit 4830ad3a3f2ccfd8cc541ed43a04fdb158ee9cab
Date:   Mon Jun 3 15:29:43 2024 -0400

    Merge pull request #53 from agrare/parse_dialog_options_as_extra_vars
    
    Parse dialog options as Extra Vars
    
    (cherry picked from commit 41c4a5c946ef88b0841bfe17eaa3e7cea393620e)
```

PR #58, opened 2024-06-06
Check embedded_terraform role in `Button::EmbeddedTerraform#disabled?`

agrare:
Prevent the user from adding credentials/repositories if the embedded_terraform role is disabled.

This is similar to what embedded_ansible's button class does https://github.com/ManageIQ/manageiq-ui-classic/blob/master/app/helpers/application_helper/button/embedded_ansible.rb

I dropped the provider check because a.) we seed embedded-terraform unconditionally and b.) there is no Provider and the EmbeddedTerraform::AutomationManager check would catch it if we didn't seed

NOTE it seems like if the entire dropdown is disabled you don't see the tooltip that says why it is disabled but that's an issue with all of these

With the role disabled:
Screenshot from 2024-06-06 13-21-02
Screenshot from 2024-06-06 13-21-26

With the role enabled:
Screenshot from 2024-06-06 13-23-57
Screenshot from 2024-06-06 13-24-08

### DIFF
--- a/app/helpers/application_helper/button/embedded_terraform.rb
+++ b/app/helpers/application_helper/button/embedded_terraform.rb
@@ -1,7 +1,10 @@
 class ApplicationHelper::Button::EmbeddedTerraform < ApplicationHelper::Button::Basic
   def disabled?
-    if Rbac.filtered(ManageIQ::Providers::EmbeddedTerraform::AutomationManager.all).empty?
+    if !MiqRegion.my_region.role_active?('embedded_terraform')
+      @error_message = _("Embedded Terraform Role is not enabled.")
+    elsif Rbac.filtered(ManageIQ::Providers::EmbeddedTerraform::AutomationManager.all).empty?
       @error_message = _("User isn't allowed to use the Embedded Terraform provider.")
     end
+    @error_message.present?
   end
 end


Fryguy:
On the embedded ansible side there's also a presence check at the end, which I think is needed: https://github.com/ManageIQ/manageiq-ui-classic/blob/master/app/helpers/application_helper/button/embedded_ansible.rb#L8C5-L8C28

agrare:
> On the embedded ansible side there's also a presence check at the end,

Oh yeah Interesting that was missing before, I think it was working because the if/elsif boolean return.  I'll add it

Fryguy:
Backported to `radjabov` in commit e929e239d4331b67bd47d3d8bc7e8075af3603c7.

```text
commit e929e239d4331b67bd47d3d8bc7e8075af3603c7
Date:   Thu Jun 6 15:10:09 2024 -0400

    Merge pull request #58 from agrare/check_embedded_terraform_role_button_disabled
    
    Check embedded_terraform role in `Button::EmbeddedTerraform#disabled?`
    
    (cherry picked from commit 56445a8f40f4d34b767b068b90fa1d17c15bd2a8)
```

PR #62, opened 2024-07-02
Adding environment variable for offline or airgap env

sudhir-kelkar:
This PR will help us add environment variables in opentofu deployment. 
We are adding 2 environment variables here
LOG4JS_LEVEL -> This will help us update log level in case we want to do further debugging
AIRGAP_OFFLINE -> This will help terraform runner to understand if customer is running in airgap environment

### DIFF
--- a/app/models/opentofu_worker.rb
+++ b/app/models/opentofu_worker.rb
@@ -74,7 +74,9 @@ class OpentofuWorker < MiqWorker
       "DATABASE_USERNAME"     => database_configuration[:username],
       "MEMCACHE_SERVERS"      => ::Settings.session.memcache_server,
       "PORT"                  => container_port,
-      "OPENTOFU_RUNNER_IMAGE" => container_image
+      "OPENTOFU_RUNNER_IMAGE" => container_image,
+      "LOG4JS_LEVEL"          => ::Settings.log.level_embedded_terraform,
+      "TF_OFFLINE"            => worker_settings[:opentofu_offline]
     }
   end
 
--- a/app/models/opentofu_worker.rb
+++ b/app/models/opentofu_worker.rb
@@ -82,7 +82,7 @@ class OpentofuWorker < MiqWorker
     env_var_array = definition[:spec][:template][:spec][:containers][0][:env]
     env_var_array.detect { |env| env[:name] == "HOME" }&.[]=(:value, "/home/node")
 
-    definition[:spec][:template][:spec][:containers][0][:env] << {:name => "LOG4JS_LEVEL", :value => :level_embedded_terraform}
+    definition[:spec][:template][:spec][:containers][0][:env] << {:name => "LOG4JS_LEVEL", :value => Settings.log.level_embedded_terraform}
     definition[:spec][:template][:spec][:containers][0][:env] << {:name => "TF_OFFLINE", :value => worker_settings[:opentofu_offline]}
 
     # these volume mounts are require by terraform runner to create the stack, mentioned it as {} so that it can be writable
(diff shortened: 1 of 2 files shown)


app/models/opentofu_worker.rb
```diff
@@ -82,6 +82,9 @@ def configure_service_worker_deployment(definition)
     env_var_array = definition[:spec][:template][:spec][:containers][0][:env]
     env_var_array.detect { |env| env[:name] == "HOME" }&.[]=(:value, "/home/node")
 
+    definition[:spec][:template][:spec][:containers][0][:env] << {:name => "LOG4JS_LEVEL", :value => "info"}
+    definition[:spec][:template][:spec][:containers][0][:env] << {:name => "AIRGAP_OFFLINE", :value => "false"}
```
bdunne, line 86: Should these be driven by settings or something? If they're going to be static values, why set them here and not in the image?

sudhir-kelkar: Customer can change based on requirement. e.g. 
- if we want more logs during issue debugging we can ask customer to change LOG4JS_LEVEL to 'debug'
- if customer is running on airgap environment, they have to change 'AIRGAP_OFFLINE' value to true

sudhir-kelkar: What do you mean by setting? Can we have such setting which can tell terraform runner deployment that customer is running on airgap environment ?

agrare: Every worker has a settings section, e.g. https://github.com/ManageIQ/manageiq-providers-embedded_terraform/blob/master/config/settings.yml#L19-L21

E.g. take a look at how we're handling the `:container_image` setting

app/models/opentofu_worker.rb
```diff
@@ -82,8 +82,8 @@ def configure_service_worker_deployment(definition)
     env_var_array = definition[:spec][:template][:spec][:containers][0][:env]
     env_var_array.detect { |env| env[:name] == "HOME" }&.[]=(:value, "/home/node")
 
-    definition[:spec][:template][:spec][:containers][0][:env] << {:name => "LOG4JS_LEVEL", :value => "info"}
-    definition[:spec][:template][:spec][:containers][0][:env] << {:name => "TF_OFFLINE", :value => "false"}
+    definition[:spec][:template][:spec][:containers][0][:env] << {:name => "LOG4JS_LEVEL", :value => :level_embedded_terraform}
```
agrare, line 85:
```suggestion
    definition[:spec][:template][:spec][:containers][0][:env] << {:name => "LOG4JS_LEVEL", :value => Settings.log.level_embedded_terraform}
```

agrare: Not sure why this is marked resolved, this is still an issue and currently will set the value to the actual symbol `:level_embedded_terraform`

agrare: Ugh github, I refreshed three times and it still shows the same diff.  Sorry.

agrare:
Okay now in order for your worker to be restarted when the settings change you should add
```ruby
self.worker_settings_paths = [
  %i[log level_embedded_terraform],
  %i[workers worker_base opentofu_worker]
]
```

agrare:
@sudhir-kelkar can you add these environment variables to the `unit_environment_variables` section as well?  That is what sets environment variables for the appliance systemd service.

agrare (approved):

Fryguy:
Backported to `radjabov` in commit 837ab532e3913231c1a926fc7eb71e29d29b6a0c.

```text
commit 837ab532e3913231c1a926fc7eb71e29d29b6a0c
Date:   Wed Jul 3 14:01:04 2024 -0400

    Merge pull request #62 from sudhir-kelkar/envOfflineAirgap
    
    Adding environment variable for offline or airgap env
    
    (cherry picked from commit 68a649b5bf4849ce210d7b7d3e90847ee48f6100)
```

PR #63, opened 2024-07-04
Marshalling boolean value to environment varaible of type string

sudhir-kelkar:
Value of opentofu_offline is set to true or false in settings.
While setting it up in deployment environment variable we need to parse it to string

### DIFF
--- a/app/models/opentofu_worker.rb
+++ b/app/models/opentofu_worker.rb
@@ -76,7 +76,7 @@ class OpentofuWorker < MiqWorker
       "PORT"                  => container_port,
       "OPENTOFU_RUNNER_IMAGE" => container_image,
       "LOG4JS_LEVEL"          => ::Settings.log.level_embedded_terraform,
-      "TF_OFFLINE"            => worker_settings[:opentofu_offline].to_s
+      "TF_OFFLINE"            => worker_settings[:opentofu_offline]
     }
   end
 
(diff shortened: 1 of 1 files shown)


app/models/opentofu_worker.rb
```diff
@@ -76,7 +76,7 @@ def unit_environment_variables
       "PORT"                  => container_port,
       "OPENTOFU_RUNNER_IMAGE" => container_image,
       "LOG4JS_LEVEL"          => ::Settings.log.level_embedded_terraform,
-      "TF_OFFLINE"            => worker_settings[:opentofu_offline]
+      "TF_OFFLINE"            => worker_settings[:opentofu_offline].to_s
```
agrare, line 79: This one isn't strictly necessary because when we write these out to the override file we are doing a string interpolation anyway

sudhir-kelkar: I think it's still good to have as:
1. It keeps the code in sync with the containerized code.
2. The Terraform runner always expects it in all lowercase, which `.to_s` already does.

sudhir-kelkar:
@miq-bot add_labels radjabov/yes

app/models/opentofu_worker.rb
sudhir-kelkar: Done

PR #71, opened 2024-09-04
Retire terraform runner stack

putmanoj:
* The Retirement(`/stack/delete` or `terraform destroy`) api from terraform-runner, needs same terraform-template, input-vars from Provision action.  Also needs **`stack_id`** created in Provision action. 
   * The config options `template_id`, TTL, default extra-vars/input-vars, etc will be same as Provision action.
   * Whereas the `stack_id` & user modified `vars` are available in Provision Job instance.  
   * When 'Retirement' action is triggered from 'My Services' page,  we need the `stack_id` & vars from Provision Job, 
   * In 'Retirement' action, we fetch Provision Job and copy `stack_id` & vars. 
 
* Separate `create_stack` & `delete_stack` methods for running Provision & Retirement actions respectively.


app/models/service_template_terraform_template.rb
```diff
@@ -37,6 +37,9 @@ def self.validate_config_info(info)
 
   def terraform_template(action)
     template_id = config_info.dig(action.downcase.to_sym, :configuration_script_payload_id)
+    if template_id.nil? && action.downcase == 'retirement'
+      template_id = config_info.dig(:provision, :configuration_script_payload_id)
```
Fryguy, line 41: Is this saying that when doing retirement use the provisioning workflow?

lib/terraform/runner.rb
```diff
@@ -18,23 +18,47 @@ def available?
...
-        Terraform::Runner::ResponseAsync.new(response.stack_id)
+        action = input_vars.key?(:action) ? input_vars[:action].downcase : ''
+        case action
+        when 'retirement'
+          #  ===== DELETE =====
+          if input_vars.key?(:stack_id) && !input_vars[:stack_id].blank?
+            stack_id = input_vars[:stack_id]
+            input_vars.delete(:action)
+            input_vars.delete(:stack_id)
+            _log.debug("Run_aysnc/delete_tack('#{stack_id}') for template: #{template_path}")
```
Fryguy, line 37:
```suggestion
            _log.debug("Run_aysnc/delete_stack('#{stack_id}') for template: #{template_path}")
```

app/models/service_template_terraform_template.rb
eworkmanoj: By default, for terraform ,  same terraform template files, vars,  but different action,  ie when 'Retirement' action -> run destroy

app/models/manageiq/providers/embedded_terraform/automation_manager/job.rb
```diff
@@ -138,4 +146,18 @@ def cleanup_git_repository
...
+      if !service_instance.nil?
+        service_instance.preprocess(action, {:extra_vars => {:miq_terraform_stack_id => stack_id}})
+        _log.debug("updated service-instance/#{service_instance.to_json} with stack_id#{stack_id}")
+      else
+        _log.info("Not found service-instance/#{service_instance_id}")
+      end
+    else
+      _log.debug("Not Provision action, no update to service-instance/#{service_instance_id} required")
+    end
+  end
```
agrare, line 162: Believe we discussed this in chat, but just to make sure this doesn't get lost I'll add here also.

I think the proper way to do this is to have the service retirement retire all of the related resources which in this case will be an OrchestrationStack.  This would also work if in the future you had a service that ends up creating multiple terraform_stacks where what you have here would preclude that.

app/models/service_terraform_template.rb
```diff
@@ -76,7 +77,25 @@ def check_refreshed(_action)
...
-    options[job_option_key(action)].deep_dup
+    case action.downcase
+    when 'retirement'
+      retire_job_options = options[job_option_key('retirement')].deep_dup
+      retire_job_options[:action] = action
+      prov_job_options = options[job_option_key('provision')].deep_dup
+      # if not already available in retirement options, (won't be available for terraform)
+      # copy extra_vars,execution_ttl,verbosity,credentials,etc from provision action
+      job_options = prov_job_options.deep_merge(retire_job_options)
+    else
```
agrare, line 88: This doesn't seem right, shouldn't we set what we need in the options for the retirement action so we don't have to do this?

app/models/service_template_terraform_template.rb
putmanoj: @agrare,  with the changes, now the template_id will be available with Retirement options also, in ServiceTemplateTerraformTemplate,

app/models/manageiq/providers/embedded_terraform/automation_manager/job.rb
putmanoj: @agrare 
The `update_service_instance_with_stack_id()` method is no longer required,  has been removed. We now fetch `stack_id` from the Provision Job,  where it available, and use it for Retirement.

Terraform-Runner as reference to all the created resources from the Provision Job,  all it needs the relevant Terraform-Tunner `stack_id` to Retire ( `terraform destroy` ) the resources.

app/models/service_terraform_template.rb
putmanoj: With new changes, we will only use `vars` from Provision Job. We still need `vars` values from Provision job, because the Retirement option only has default value in ServiceTemplates the default values, we want the user-modified values from the Provision Job.

lib/terraform/runner.rb
```diff
@@ -9,32 +9,59 @@ class << self
       def available?
         return @available if defined?(@available)
 
-        response = terraform_runner_client.get('ping')
-        @available = response.status == 200
+        response = terraform_runner_client.get('live')
+        @available = response.status == 200 && JSON.parse(response.body)['status'] == 'UP'
```
agrare, line 13: @putmanoj change looks unrelated to the retirement work, can you pull it out into its own PR?

app/models/service_terraform_template.rb
```diff
@@ -75,8 +76,40 @@ def check_refreshed(_action)
...
   def get_job_options(action)
-    options[job_option_key(action)].deep_dup
+    job_options = options[job_option_key(action)].deep_dup
+
+    # Add additional vars,
+    # :miq_action                    -> required for Retirement action
+    # :miq_terraform_runner_stack_id -> Terraform-Runner stack_id required by Retirement action.
+
+    case action
+    when ResourceAction::RETIREMENT
```
agrare, line 94: The `case` with only one state is odd, could you just do `if action == ResourceAction::RETIREMENT`

app/models/service_terraform_template.rb
```diff
@@ -75,8 +76,40 @@ def check_refreshed(_action)
 
   private
 
+  def job(action)
+    stack(action).miq_task.job
+  rescue err
```
agrare, line 81: What exception are you expecting to catch here?  If it is just e.g. miq_task being `nil` then you're probably better off with `stack(action)&.miq_task&.job` without the exception handler+logging

app/models/service_terraform_template.rb
```diff
@@ -75,8 +76,40 @@ def check_refreshed(_action)
...
+    case action
+    when ResourceAction::RETIREMENT
+      prov_job = job(ResourceAction::PROVISION)
+      if prov_job.present? && prov_job.options.present?
+        # Copy input-vars from Provision(terraform apply) action,
+        # the Retirement(terraform destroy) action will use same input-vars/values.
+        if prov_job.options.key?(:input_vars) && prov_job.options[:input_vars].key?(:extra_vars)
+          prov_vars = prov_job.options[:input_vars][:extra_vars].deep_dup
+          job_options[:extra_vars] = prov_vars.deep_merge(job_options[:extra_vars])
+        end
```
agrare, line 102: Could this whole thing become?

```
prov_vars = prov_job.options.dig(:input_vars, :extra_vars)
job_options[:extra_vars].deep_merge!(prov_vars) if prov_vars
```

lib/terraform/runner.rb
```diff
@@ -9,32 +9,59 @@ class << self
...
-          :env_vars    => env_vars
-        )
-        Terraform::Runner::ResponseAsync.new(response.stack_id)
+        action   = input_vars[:miq_action]
+        stack_id = input_vars[:miq_terraform_runner_stack_id]
+        input_vars.delete(:miq_action)
+        input_vars.delete(:miq_terraform_runner_stack_id)
+
+        case action
+        when ResourceAction::RETIREMENT
```
agrare, line 37: This doesn't seem right, I expected that we'd expose the create_stack and delete_stack on Terraform::Runner rather than using some hash value in one of the arguments to completely change the behavior of the method call.

lib/terraform/runner.rb
putmanoj: This PR was created long back, yes will move to another PR

lib/terraform/runner.rb
putmanoj: Originally, my thoughts were also this, but this made things easier

app/models/manageiq/providers/embedded_terraform/automation_manager/job.rb
```diff
@@ -21,12 +23,28 @@ def pre_execute
...
-
-    response = Terraform::Runner.run(decrypt_extra_vars(extra_vars), template_path, :credentials => credentials, :env_vars => options[:env_vars])
-
-    options[:terraform_stack_id] = response.stack_id
-    save!
+    input_vars    = options.dig(:input_vars, :extra_vars) || {}
+    action        = options[:action]
+
+    case action
+    when ResourceAction::RETIREMENT
```
agrare, line 30: This looks like it could easily just be
```suggestion
    if action == ResourceAction::RETIREMENT
```

I don't see the case buying you anything

app/models/manageiq/providers/embedded_terraform/automation_manager/job.rb
```diff
@@ -106,8 +124,8 @@ def stack_response
     @stack_response ||= Terraform::Runner::ResponseAsync.new(options[:terraform_stack_id])
   end
 
-  def decrypt_extra_vars(extra_vars)
-    result = extra_vars.deep_dup
+  def decrypt_input_vars(input_vars)
```
agrare, line 127: Maybe just rename this `decrypt_vars` since it could work for extra_vars or input_vars?  It doesn't depend on anything other than the argument you pass in.

app/models/manageiq/providers/embedded_terraform/automation_manager/job.rb
```diff
@@ -106,8 +124,8 @@ def stack_response
...
+    result = input_vars.deep_dup
     result.transform_values! { |val| val.kind_of?(String) ? ManageIQ::Password.try_decrypt(val) : val }
```
agrare, line 129: @putmanoj could you do `input_vars.transform_values { |val| val.kind_of?(String) ? ManageIQ::Password.try_decrypt(val) : val }` instead of `deep_dup` plus `transform_values!` ?

`transform_values!` means you're going out of your way to change the original object not create a copy, but it looks like you _do_ want to create a copy and not modify the original

lib/terraform/runner.rb
```diff
@@ -9,39 +9,73 @@ class << self
...
+      alias create_stack run_async
+
+      # Delete(destroy) terraform-runner created stack resources.
+      def delete_stack(stack_id, template_path, input_vars: {}, credentials: [], env_vars: {})
+        run_async(template_path,
+                  :input_vars  => input_vars,
+                  :credentials => credentials,
+                  :env_vars    => env_vars,
+                  :action      => ResourceAction::RETIREMENT,
+                  :stack_id    => stack_id)
```
agrare, line 75: What does calling run_async which then calls `delete_stack_job` buy us?  Could you just call `        Terraform::Runner::ResponseAsync.new(delete_stack_job(...))` directly?

putmanoj: We originally had run/run_async,  so kept it as is for now..

putmanoj: we can do this change in next PR

putmanoj: done, pushed the changes

lib/terraform/runner.rb
putmanoj: done

app/models/manageiq/providers/embedded_terraform/automation_manager/job.rb
```diff
@@ -106,9 +124,8 @@ def stack_response
     @stack_response ||= Terraform::Runner::ResponseAsync.new(options[:terraform_stack_id])
   end
 
-  def decrypt_extra_vars(extra_vars)
-    result = extra_vars.deep_dup
-    result.transform_values! { |val| val.kind_of?(String) ? ManageIQ::Password.try_decrypt(val) : val }
+  def decrypt_vars(input_vars)
+    input_vars.transform_values! { |val| val.kind_of?(String) ? ManageIQ::Password.try_decrypt(val) : val }
```
agrare, line 128: I thought you wanted to not modify the argument that was passed in, if that is still the case you want `transform_values` instead of `transform_values!` (the difference is the latter modifies the argument that is passed in where the latter creates a new object)

agrare:
This looks okay to merge once the issue https://github.com/ManageIQ/manageiq-providers-embedded_terraform/pull/71#discussion_r1834520758 is resolved, we can drop the run_async method in a follow-up or in this PR if you want.

agrare (approved):

Fryguy:
Backported to `radjabov` in commit 1d9696d65ec11ee0c7819639c1228b094c58470b.

```text
commit 1d9696d65ec11ee0c7819639c1228b094c58470b
Date:   Fri Nov 8 12:36:51 2024 -0500

    Merge pull request #71 from putmanoj/retire-tf-runner-stack
    
    Retire terraform runner stack
    
    (cherry picked from commit 75c8d4417b829bdd1bf92727696fd3dfea94b3f8)
```

PR #72, opened 2024-09-09
[WIP] To Show terraform output in Services

jaisejose1123:
**PR raised for Show terraform output in Services**
[https://jsw.ibm.com/browse/CP4AIOPS-1082](url)

**Before**
After executing Terraform through ManageIQ, the results were not stored in the MiqTask and Job tables.

image
The context_data field in the MiqTask table was empty.

image
The context field in the Job table was empty.

image
start_time of OrchestrationStack was nill

**After**
Now, Terraform results are being stored in the MiqTask and Job tables.

image
The context_data field stores the Terraform results in the MiqTask table.

image
The context field stores the Terraform results in the MiqTask table.

image
start_time of OrchestrationStack get updated

### DIFF
--- a/app/models/manageiq/providers/embedded_terraform/automation_manager/job.rb
+++ b/app/models/manageiq/providers/embedded_terraform/automation_manager/job.rb
@@ -15,7 +15,7 @@ class ManageIQ::Providers::EmbeddedTerraform::AutomationManager::Job < Job
 
   def pre_execute
     checkout_git_repository
-    signal(:execute)
+    route_signal(:execute)
   end
 
   def execute
@@ -27,26 +27,63 @@ class ManageIQ::Providers::EmbeddedTerraform::AutomationManager::Job < Job
 
     options[:terraform_stack_id] = response.stack_id
     save!
-
-    queue_poll_runner
+    route_signal(:poll_runner)
+    #queue_poll_runner
   end
 
   def poll_runner
     if running?
       queue_poll_runner
     else
-      signal(:post_execute)
+      logs_update
+      route_signal(:post_execute)
     end
   end
 
   def post_execute
-    cleanup_git_repository
-
+    
+    #cleanup_git_repository
+    
     return queue_signal(:finish, message, status) if success?
-
+    #queue_signal(:finish, message, status)
+    return queue_signal(:abort, "Failed to run ansible", "error") unless success?
     _log.error("Failed to run template: [#{error_message}]")
-
+    
     abort_job("Failed to run template", "error")
+  end
+
+  def logs_update
+    terraform_response = {
+      stack_id: stack_response&.response&.stack_id,
+      stack_name: stack_response&.response&.stack_name,
+      status: stack_response&.response&.status,
+      stack_job_end_time: stack_response&.response&.stack_job_end_time,
+      stack_job_start_time: stack_response&.response&.stack_job_start_time,
+      created_at: stack_response&.response&.created_at,
+      message: stack_response&.response&.message,
+      error_message: stack_response&.response&.error_message
+    }
+
+    miq_task.context_data ||= {}
+    miq_task.context_data[:terraform_response] = terraform_response
+    context[:terraform_response] = terraform_response
+    
+    update!(:context => context, :started_on => context[:terraform_response][:stack_job_start_time])
+    miq_task.update(context_data: miq_task.context_data)
+  end
+
+  def queue_signal(*args, deliver_on: nil, msg_timeout: nil)
+    role     = options[:role] || "ems_operations"
+    priority = options[:priority] || MiqQueue::NORMAL_PRIORITY
+    super(*args, :msg_timeout => msg_timeout, :priority => priority, :role => role, :deliver_on => deliver_on, :server_guid => MiqServer.my_server.guid)
+  end
+
+  def route_signal(*args, deliver_on: nil)
+    if MiqEnvironment::Command.is_podified?
+      signal(*args)
+    else
+      queue_signal(*args, :deliver_on => deliver_on)
+    end
   end
 
   alias initializing dispatch_start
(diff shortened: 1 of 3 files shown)


app/models/manageiq/providers/embedded_terraform/automation_manager/job.rb
```diff
@@ -40,15 +40,35 @@ def poll_runner
   end
 
   def post_execute
-    cleanup_git_repository
-
-    return queue_signal(:finish, message, status) if success?
+    if success?
```
agrare, line 43: @jaisejose1123 is it possible to get logs on failure also?  I think that's almost more important in order to figure out what went wrong.

app/models/manageiq/providers/embedded_terraform/automation_manager/job.rb
```diff
@@ -40,15 +40,35 @@ def poll_runner
...
+      stack_name: stack_response&.response&.stack_name,
+      status: stack_response&.response&.status,
+      stack_job_end_time: stack_response&.response&.stack_job_end_time,
+      stack_job_start_time: stack_response&.response&.stack_job_start_time,
+      created_at: stack_response&.response&.created_at,
+      message: stack_response&.response&.message
+    }
+
+    miq_task.context_data ||= {}
+    miq_task.context_data[:terraform_respons] = terraform_response
```
agrare, line 65: Typo, should be `:terraform_response`

app/models/manageiq/providers/embedded_terraform/automation_manager/job.rb
```diff
@@ -40,15 +40,35 @@ def poll_runner
...
+    context[:terraform_respons] = terraform_response
```
agrare, line 66: Same here, `:terraform_response`

app/models/manageiq/providers/embedded_terraform/automation_manager/job.rb
```diff
@@ -40,15 +40,35 @@ def poll_runner
...
+    
+    update!(:context => context, :started_on => context[:terraform_respons][:stack_job_start_time])
```
agrare, line 68: Do we want to override the started_on property?  Was this not being set before?  If it wasn't being set we probably want to set it earlier than `post_execute`

agrare:
@jaisejose1123 are you able to see the log output on the stack or do you have to implement raw_stdout like EmbeddedAnsible does? https://github.com/ManageIQ/manageiq/blob/master/app/models/manageiq/providers/embedded_ansible/automation_manager/job.rb#L149-L151

app/models/manageiq/providers/embedded_terraform/automation_manager/job.rb
agrare: E.g. embedded_ansible sets this on execute https://github.com/ManageIQ/manageiq/blob/master/app/models/manageiq/providers/ansible_runner_workflow.rb#L50-L52

jaisejose1123:
> @jaisejose1123 are you able to see the log output on the stack or do you have to implement raw_stdout like EmbeddedAnsible does? https://github.com/ManageIQ/manageiq/blob/master/app/models/manageiq/providers/embedded_ansible/automation_manager/job.rb#L149-L151

@agrare We are storing the logs in the database using the format shown above, so we can directly query them from the database.
image

jaisejose1123:
> > @jaisejose1123 are you able to see the log output on the stack or do you have to implement raw_stdout like EmbeddedAnsible does? https://github.com/ManageIQ/manageiq/blob/master/app/models/manageiq/providers/embedded_ansible/automation_manager/job.rb#L149-L151
> 
> @agrare We are storing the logs in the database using the format shown below, so we can directly query them from the database. image

The result messages are stored in a human-readable format, allowing us to directly access the result by calling the key from the hash.

agrare:
@jaisejose1123 can you show an example of where this is visible on the UI?  I'm surprised we don't have to implement OrchestrationStack stdout like embedded ansible does.

jaisejose1123:
> backend code that uses TerminalToHtml

 completed the UI work but am currently fixing a few minor errors. The backend code using TerminalToHtml is already finished. I’ll update the PR soon.

app/models/manageiq/providers/embedded_terraform/automation_manager/job.rb
```diff
@@ -27,28 +27,65 @@ def execute
...
+  def queue_signal(*args, deliver_on: nil, msg_timeout: nil)
+    role     = options[:role] || "ems_operations"
+    priority = options[:priority] || MiqQueue::NORMAL_PRIORITY
+    super(*args, :msg_timeout => msg_timeout, :priority => priority, :role => role, :deliver_on => deliver_on, :server_guid => MiqServer.my_server.guid)
+  end
+
+  def route_signal(*args, deliver_on: nil)
+    if MiqEnvironment::Command.is_podified?
+      signal(*args)
+    else
```
agrare, line 84: @jaisejose1123 do we need this for terraform?  The reason it is like this in embedded_ansible is because the files/directories related to the ansible-runner execution are literally on the pod's filesystem so we have to run it on the same pod, but this has a number of drawbacks and something we want to get away from.

With the opentofu-runner any pod in the namespace should be able to hit that API so any pod could run the next state.

app/models/manageiq/providers/embedded_terraform/automation_manager/job.rb
```diff
@@ -27,28 +27,65 @@ def execute
...
+      logs_update
+      route_signal(:post_execute)
     end
   end
 
   def post_execute
-    cleanup_git_repository
-
+    
+    #cleanup_git_repository
```
agrare, line 45: Are we no longer cleaning up the git repository?

putmanoj:
We can close this PR,  has been replaced by https://github.com/ManageIQ/manageiq-providers-embedded_terraform/pull/78

PR #73, opened 2024-09-18
OpentofuWorker#stop_worker should delete the pod

agrare:
OpentofuWorker is a ServiceWorker but also isn't scalable (aka it is a deployment-per-worker type).

This means that when we want to stop an OpentofuWorker we have to delete the deployment where previously it would scale the deployment to 1 and not actually delete the pod.

### DIFF
--- a/app/models/opentofu_worker.rb
+++ b/app/models/opentofu_worker.rb
@@ -49,6 +49,13 @@ class OpentofuWorker < MiqWorker
     }
   end
 
+  # The opentofu-runner pod is a service worker but isn't scalable so the
+  # stop_container method should delete_container_objects like a deployment-per-worker
+  # not scale_deployment like a service_worker.
+  def stop_container
+    delete_container_objects
+  end
+
   private
 
   # There can only be a single instance running so the unit name can just be


Fryguy:
@bdunne Please review. 

I am curious why this isn't the default on the base class.

Fryguy:
@agrare are we trying to restart the worker in this case or actually turn it off completely (like if someone disabled the role)?

agrare:
> I am curious why this isn't the default on the base class.

This worker includes the ServiceWorker mixin (https://github.com/ManageIQ/manageiq/blob/master/app/models/miq_worker/service_worker.rb) since it's an API, but it also isn't scalable which is a Deployment per Worker (https://github.com/ManageIQ/manageiq/blob/master/app/models/miq_worker/deployment_per_worker.rb) so this is kind of a hybrid.

agrare:
> re we trying to restart the worker in this case or actually turn it off completely (like if someone disabled the role)?

Both, to restart a worker we stop it and then the sync starts a new one.

Fryguy:
> Both, to restart a worker we stop it and then the sync starts a new one.

Oh that's strange - I would expect it to just kill the pod, and then kube will just start a new one.

Disabling a role makes sense to me that you'd kill the delpoyment, but during restarting a worker feels like overkill to do that (I mean it will work, but it's not normally how you'd do it if you were manually operating the cluster).

agrare:
> Oh that's strange - I would expect it to just kill the pod, and then kube will just start a new one.

Well currently it does neither of those for this worker :laughing:  We can definitely try to improve that in core but that feels out of scope of this fix

agrare:
`MiqWorker#restart`: https://github.com/ManageIQ/manageiq/blame/master/app/models/miq_worker.rb#L407-L408
So restart has been an alias for stop for a long time, it probably makes sense when miq_server worker monitor was _the_ service manager, less so with systemd/kubernetes

agrare:
Actually @Fryguy we have to delete the deployment if we're changing settings because the deployment has the spec including the env vars so just restarting the pod won't pick up new settings.

jrafanie:
> Actually @Fryguy we have to delete the deployment if we're changing settings because the deployment has the spec including the env vars so just restarting the pod won't pick up new settings.

Note, we [patch the deployment](https://github.com/ManageIQ/manageiq/blob/master/app/models/miq_server/worker_management/kubernetes.rb#L109-L117) to update requests/limits changes on cpu and memory.  That's very specific so maybe that's what you mean, we need to manually sync specific settings or let it be recreated correctly after removing it.

agrare:
@jrafanie actually no I'm talking specifically about these two environment variables that are being pulled from settings https://github.com/ManageIQ/manageiq-providers-embedded_terraform/blob/b91bddc2c98fc2967c7cb1273c8686bed70aefa5/app/models/opentofu_worker.rb#L103-L104

Since opentofu-runner isn't a rails worker it gets all settings from env vars not `::Settings`

jrafanie:
> Since opentofu-runner isn't a rails worker it gets all settings from env vars not `::Settings`, we have code in core to handle settings changes with non-rails workers and it restarts the workers to pick up the new settings (the worker declares what settings keys it cares about)

Ok, good to know.  Thanks for the explanation.

Fryguy:
> we have to delete the deployment if we're changing settings because the deployment has the spec including the env vars so just restarting the pod won't pick up new settings.

Fair point. :+1:

> Note, we [patch the deployment](https://github.com/ManageIQ/manageiq/blob/master/app/models/miq_server/worker_management/kubernetes.rb#L109-L117) to update requests/limits changes on cpu and memory.

Yeah, in that case because the deployment is modified it's retaining certain properties, and kube takes care of "restarting" the worker (really it kills the pod, then realizes the pod is missing and creates a new one from the updated deployment). I guess "restart" can have multiple nuanced meanings depending on if settings are changing or not.

> Well currently it does neither of those for this worker 😆 We can definitely try to improve that in core but that feels out of scope of this fix

Yeah my intention isn't to derail this PR cause it fixes a real bug, and I'm good with the PR. I more wanted to bring up that moving forward we may want to consider other ways to restart then killing the deployment (such as patching the deployment or killing the pod).

Fryguy (approved):
LGTM, but I really want @bdunne to review.

jrafanie (approved):
LGTM.  Let's go @bdunne!

Fryguy:
Backported to `radjabov` in commit ed91a35d7de245dc67f10ddcc42d038cea8b357c.

```text
commit ed91a35d7de245dc67f10ddcc42d038cea8b357c
Author: Brandon Dunne <brandondunne@hotmail.com>
Date:   Tue Sep 24 17:30:33 2024 -0400

    Merge pull request #73 from agrare/opentofu_runner_stop_should_delete_deployment
    
    OpentofuWorker#stop_worker should delete the pod
    
    (cherry picked from commit 0fd9aa047410b5345e34ee1787aaec6d9d8c3b43)
```

PR #75, opened 2024-09-23
Parse terraform template & create new service dialog with parsed input vars

putmanoj:
* Parse terraform template, while syncing the git repository
* Send terraform template to TerraformRunner service, parse for input/output vars
* In Service Template item, use this parsed input-vars to create textboxes for new Service Dialog.


app/models/dialog/terraform_template_service_dialog.rb
```diff
@@ -0,0 +1,92 @@
+require "JSON"
```
agrare, line 1: I think you mean `require "json"`
```
>> require 'JSON'
<internal:/usr/lib/ruby/vendor_ruby/rubygems/core_ext/kernel_require.rb>:86:in `require': cannot load such file -- JSON (LoadError)
```

Also for delay load purposes (keeping memory usage low) we prefer to put the require in the method/class that it is used in.  In this case you can add it right above here https://github.com/ManageIQ/manageiq-providers-embedded_terraform/pull/75/files#diff-153c36f8373e1ea8b9918d4a35cdd505014f31e78f92c7702ffd53eb32892e27R26

app/models/dialog/terraform_template_service_dialog.rb
```diff
@@ -0,0 +1,92 @@
...
+        if extra_vars.present?
+          add_variables_group(tab, 1, extra_vars)
+        end
+        dialog.save!
+      end
+    end
+
+    private
+
+    def add_template_variables_group(tab, position, terraform_template)
```
agrare, line 25: Is position necessary as an argument?  This only ever seems to be called with `position=0`, likewise `add_variables_group` only ever called with `position=1`

app/models/dialog/terraform_template_service_dialog.rb
```diff
@@ -0,0 +1,92 @@
...
+          description = var_info["description"]
+          if description.blank?
+            description = key
+          end
+
+          # TODO: use these when adding variable field
+          # type = var_info["type"]
+          # secured = var_info["secured"]
+
+          if hidden == true
```
agrare, line 53: The explicit `== true` isn't required

```suggestion
          if hidden
```

app/models/dialog/terraform_template_service_dialog.rb
```diff
@@ -0,0 +1,92 @@
...
+            _log.info("Not adding text-box for hidden variable: #{key}")
```
agrare, line 54: This is probably not a helpful log message, you could do
```
add_variable_field(key, value, dialog_group, index, label, description, required, readonly) unless hidden
```

app/models/dialog/terraform_template_service_dialog.rb
```diff
@@ -0,0 +1,92 @@
...
+          value = var_info["default"]
+          value = value.to_json if [Hash, Array].include?(value.class)
+          required = var_info["required"]
+          readonly = var_info["immutable"]
+          hidden = var_info["hidden"]
+          label = var_info["label"]
+          description = var_info["description"]
+          if description.blank?
+            description = key
+          end
```
agrare, line 47: This is a good time to one-line this
```suggestion
          description = key if description.blank?
```

app/models/manageiq/providers/embedded_terraform/automation_manager/configuration_script_source.rb
```diff
@@ -75,8 +75,11 @@ def self.template_name_from_git_repo_url(git_repo_url, relative_path)
   def find_templates_in_git_repo
     template_dirs = {}
 
+    # checkout repo, for sending files to terraform-runner to parse for input/ouput vars.
+    git_checkout_tempdir = checkout_git_repo
+
     # traverse through files in git-worktree
-    git_repository.update_repo
+    # git_repository.update_repo - already done through checkout_git_repo
```
agrare, line 82: We shouldn't commit commented out code
```suggestion
```

app/models/manageiq/providers/embedded_terraform/automation_manager/configuration_script_source.rb
```diff
@@ -85,22 +88,51 @@ def find_templates_in_git_repo
               .group_by          { |file| File.dirname(file) }
               .select            { |_dir, files| files.any? { |f| f.end_with?(".tf", ".tf.json") } }
               .transform_values! { |files| files.map { |f| File.basename(f) } }
-              .each do |parent_dir, files|
-        name = self.class.template_name_from_git_repo_url(git_repository.url, parent_dir)
+              .each do |relative_path, files|
+        name = self.class.template_name_from_git_repo_url(git_repository.url, relative_path)
+
+        template_full_path = File.join(git_checkout_tempdir, relative_path)
```
agrare, line 94: I'm pretty sure git_checkout_tempdir is a Pathname (please verify) so you can simply `git_checkout_tempdir.join(relative_path)`

lib/terraform/runner.rb
```diff
@@ -180,6 +187,26 @@ def encoded_zip_from_directory(template_path)
...
+
+        # TODO: use tags,env_vars
+        payload = {
+          :templateZipFile => encoded_zip_file,
+        }
+
+        http_response = terraform_runner_client.post(
+          "api/template/variables",
+          *json_post_arguments(payload)
+        )
```
agrare, line 205: Interesting I thought parsing the template variables was something we could do entirely from the committed files, didn't realize it was an API call to the runner

app/models/service_template_terraform_template.rb
```diff
@@ -17,6 +17,9 @@ def self.create_catalog_item(options, _auth_user)
 
     transaction do
       create_from_options(options).tap do |service_template|
+        dialog_ids = service_template.send(:create_dialogs, config_info)
+        config_info.deep_merge!(dialog_ids)
+        service_template.options[:config_info].deep_merge!(dialog_ids)
```
agrare, line 22: Did you mean to do
```suggestion
        config_info.deep_merge!(dialog_ids)
        service_template.options[:config_info] = config_info
```

app/models/service_template_terraform_template.rb
```diff
@@ -41,4 +44,23 @@ def terraform_template(action)
 
     ManageIQ::Providers::EmbeddedTerraform::AutomationManager::Template.find(template_id)
   end
+
+  def create_dialogs(config_info)
+    # [:provision, :retirement, :reconfigure].each_with_object({}) do |action, hash|
+    [:provision,].each_with_object({}) do |action, hash|
```
agrare, line 50: No sense in looping over a single key, also best not to commit commented out code

app/models/service_template_terraform_template.rb
```diff
@@ -41,4 +44,23 @@ def terraform_template(action)
...
+      new_dialog_name = info.key?(:new_dialog_name) ? info[:new_dialog_name] : "Dialog-#{template_name}"
+
+      hash[action] = {:dialog_id => create_new_dialog(new_dialog_name, terraform_template(action), info[:extra_vars]).id}
+    end
+  end
+
+  def create_new_dialog(dialog_name, terraform_template, extra_vars)
+    Dialog::TerraformTemplateServiceDialog.create_dialog(dialog_name, terraform_template, extra_vars)
+  end
+  private :create_new_dialog
```
agrare, line 65:
```suggestion

  private
  
  def create_new_dialog(dialog_name, terraform_template, extra_vars)
    Dialog::TerraformTemplateServiceDialog.create_dialog(dialog_name, terraform_template, extra_vars)
  end
```

lib/terraform/runner/variables_response.rb
```diff
@@ -0,0 +1,29 @@
...
+
+      # @return [String] Extracted attributes from the JSON response body object
+      def self.parsed_response(http_response)
+        data = JSON.parse(http_response.body)
+        _log.debug("data : #{data}")
+        Terraform::Runner::VariablesResponse.new(
+          :template_input_params  => data['template_input_params'],
+          :template_output_params => data['template_output_params'],
+          :terraform_version      => data['terraform_version']
+        )
```
agrare, line 19: Do we need to create a new response class when we could just return the payload from the api call?  It doesn't look like there is much translation being done other than string to symbol keys :shrug:

spec/lib/terraform/runner_spec.rb
```diff
@@ -352,4 +352,48 @@ def verify_req(req)
...
+        expect(template_variables_stub).to(have_been_requested.times(1))
+
+        expect(response.template_input_params.length).to(eq(1))
+        input_var = response.template_input_params[0]
+        expect(input_var["name"]).to(eq("name"))
+        expect(input_var["label"]).to(eq("name"))
+        expect(input_var["type"]).to(eq("string"))
+        expect(input_var["default"]).to(eq("World"))
+        expect(input_var["required"]).to(eq(true))
+        expect(input_var["secured"]).to(eq(false))
```
agrare, line 389:
```suggestion
        expect(response.template_input_params.first).to have_attributes(
          "name" => "name",
          "label" => "name",
          etc...
```

putmanoj: now using include() to match hash

app/models/dialog/terraform_template_service_dialog.rb
putmanoj: Improved the code,  though not removing position argument itself.
the first group always start at '0',  
next group will may start at '1', if there are terraform template variables 
else next group will start at '0', if there are NO terraform template variables

lib/terraform/runner.rb
putmanoj: The best way to parse terraform template is to use hashicorp's own tool => [terraform-config-inspect](https://github.com/hashicorp/terraform-config-inspect).  

The same is already used in CAM,  so using the api.

app/models/manageiq/providers/embedded_terraform/automation_manager/configuration_script_source.rb
putmanoj:
```
undefined method `join' for "/var/folders/0j/t_rw5py52bl3yd8sx1bzd4280000gn/T/embedded-terraform-runner-git20240924-65302-v3ebjh":String

        template_full_path = git_checkout_tempdir.join(relative_path)
        
```
Can't use

app/models/service_template_terraform_template.rb
putmanoj: Though still using the loop, as latter we will have`:retirement, :reconfiguration`.

But then you are right, we may not need `:retirement, :reconfiguration`,   as currently use case for terraform is to use the same template,  with same dialog for both `:retirement, :reconfiguration`.

putmanoj: Made the changes to create new dialog for :provision action,  and use the same dialog_id with both :retirement and :reconfiguration action.

app/models/service_template_terraform_template.rb
agrare: Hey @putmanoj this is I think the last issue outstanding before I can merge

app/models/manageiq/providers/embedded_terraform/automation_manager/configuration_script_source.rb
```diff
@@ -85,22 +87,51 @@ def find_templates_in_git_repo
...
+  end
+
+  # checkout git repo to temp dir
+  def checkout_git_repo
+    git_checkout_tempdir = Dir.mktmpdir("embedded-terraform-runner-git")
+
+    _log.debug("Checking out git repository to #{git_checkout_tempdir}...")
+    checkout_git_repository(git_checkout_tempdir)
+    git_checkout_tempdir
+  end
```
agrare, line 128: I'm surprised there isn't a way to do this in the base `EmbeddedAutomationManager::ConfigurationScript` class, ideally we could have a `with_git_repository` which took a block and automatically handled the tempdir creation and cleanup.

putmanoj: ok,  let me do this another PR,  rather than is PR.

app/models/service_template_terraform_template.rb
putmanoj: I did see your comments,  as this code is 'copy' from ansible, kept this for last for figure if above can done.  Will check today

putmanoj: Done

app/models/manageiq/providers/embedded_terraform/automation_manager/configuration_script_source.rb
```diff
@@ -85,22 +87,51 @@ def find_templates_in_git_repo
...
+          :output_vars       => output_vars,
+          :terraform_version => terraform_version,
         }
       end
     end
 
     template_dirs
+  rescue => error
+    _log.error("Failing scaning for terraform templates in the git repo: #{error}")
+    raise error
```
agrare, line 110: Minor, but you don't need to `raise error` to re-raise the same exception in a rescue you can simply `raise` and it'll reraise the same error (won't hold up this PR for this though)

spec/lib/terraform/runner_spec.rb
```diff
@@ -352,4 +352,60 @@ def verify_req(req)
...
+    describe '.parse_template_variables input/output vars' do
+      template_variables_stub = nil
+
+      def verify_req(req)
+        body = JSON.parse(req.body)
+        expect(body).to(have_key('templateZipFile'))
+      end
+
+      before do
+        ENV["TERRAFORM_RUNNER_URL"] = "https://1.2.3.4:7000"
```
agrare, line 366: This can cause issues if other specs need different values here in the future, because the before is setting a real ENV var but nothing is resetting it.

This is where `stub_const` helps, e.g.:
`before { stub_const("ENV", ENV.to_h.merge("CONTAINER_IMAGE_NAMESPACE" => "registry.example.com/manageiq")) }`

spec/lib/terraform/runner_spec.rb
```diff
@@ -352,4 +352,60 @@ def verify_req(req)
...
+
+        @hello_world_variables_response = JSON.parse(File.read(File.join(__dir__, "runner/data/responses/hello-world-variables-success.json")))
```
agrare, line 368: Why is this an instance var?  It looks like it is only used here https://github.com/ManageIQ/manageiq-providers-embedded_terraform/pull/75/files#diff-3368830fa87ae1d0043c2385e9f9e0276e1e352e87c3fced3ca229203f1f6327R374

spec/models/manageiq/providers/embedded_terraform/automation_manager/configuration_script_source_spec.rb
```diff
@@ -15,7 +15,21 @@
...
     let(:repo_dir_structure) { %w[hello_world.tf] }
 
+    def verify_req(req)
+      body = JSON.parse(req.body)
+      expect(body).to(have_key('templateZipFile'))
+    end
+
     before do
+      ENV["TERRAFORM_RUNNER_URL"] = "https://1.2.3.4:7000"
+      @hello_world_variables_response = JSON.parse(File.read(File.join(__dir__, "../../../../../lib/terraform/runner/data/responses/hello-world-variables-success.json")))
```
agrare, line 25: I think you want `File.join(__dir__, "runner/data/responses/hello-world-variables-success.json")` here also

spec/models/manageiq/providers/embedded_terraform/automation_manager/configuration_script_source_spec.rb
```diff
@@ -15,7 +15,21 @@
     let(:repos)              { Dir.glob(File.join(repo_dir, "*")) }
     let(:repo_dir_structure) { %w[hello_world.tf] }
 
+    def verify_req(req)
+      body = JSON.parse(req.body)
+      expect(body).to(have_key('templateZipFile'))
+    end
+
     before do
+      ENV["TERRAFORM_RUNNER_URL"] = "https://1.2.3.4:7000"
```
agrare, line 24: Same comment re: `stub_const`

spec/models/manageiq/providers/embedded_terraform/automation_manager/configuration_script_source_spec.rb
```diff
@@ -15,7 +15,21 @@
...
+      @hello_world_variables_response = JSON.parse(File.read(File.join(__dir__, "../../../../../lib/terraform/runner/data/responses/hello-world-variables-success.json")))
+      @template_variables_stub = stub_request(:post, "https://1.2.3.4:7000/api/template/variables")
```
agrare, line 26: Is this `@template_variables_stub` used anywhere?

spec/models/manageiq/providers/embedded_terraform/automation_manager/configuration_script_source_spec.rb
```diff
@@ -85,10 +99,11 @@ def files_in_repository(git_repo_dir)
         expect(names.first).to(eq("/hello_world_local"))
 
         expected_hash = {
-          "relative_path" => File.dirname(*repo_dir_structure),
-          "files"         => [File.basename(*repo_dir_structure)],
-          "input_vars"    => nil,
-          "output_vars"   => nil
+          "relative_path"     => File.dirname(*repo_dir_structure),
+          "files"             => [File.basename(*repo_dir_structure)],
+          "input_vars"        => @hello_world_variables_response['template_input_params'],
```
agrare, line 104: If you need to use `@hello_world_variables_response` recommend using `let()` instead of an instance-variable:
```
let(:hello_world_response) { JSON.parse(File.read(File.join(__dir__, "runner/data/responses/hello-world-variables-success.json"))) }
```

Then you can just do this here:
```suggestion
          "input_vars"        => hello_world_response['template_input_params'],
```

spec/lib/terraform/runner_spec.rb
putmanoj: :-| frankly did not know, you could stub ENV var this way in ruby ...
 
 Thanks, will update the code accordingly

agrare: :tada: for learning something new

spec/models/manageiq/providers/embedded_terraform/automation_manager/configuration_script_source_spec.rb
putmanoj: Won't do,  different subdirectories in the parent directory tree,
Here only reusing already available data file from another spec

```
      Failure/Error: @hello_world_variables_response = JSON.parse(File.read(File.join(__dir__, "runner/data/responses/hello-world-variables-success.json")))
      
      Errno::ENOENT:
        No such file or directory @ rb_sysopen - /Users/manoj/git/ManageIQ/manageiq-providers-embedded_terraform/spec/models/manageiq/providers/embedded_terraform/automation_manager/runner/data/responses/hello-world-variables-success.json
      # ./spec/models/manageiq/providers/embedded_terraform/automation_manager/configuration_script_source_spec.rb:26:in `read'
```

spec/models/manageiq/providers/embedded_terraform/automation_manager/configuration_script_source_spec.rb
```diff
@@ -1,3 +1,5 @@
+TERRAFORM_RUNNER_URL = 'https://1.2.3.4:7000'.freeze
```
agrare, line 1: We shouldn't add constants to the global scope, a better way to do this is to have a `let(:terraform_runner_url) { 'https://1.2.3.4:7000' }` inside your `RSpec.describe` block so it is available to every test in this file but not globally.

spec/lib/terraform/runner_spec.rb
```diff
@@ -1,12 +1,15 @@
 require 'webmock/rspec'
 require 'json'
 
+TERRAFORM_RUNNER_URL = "https://1.2.3.4:7000".freeze
```
agrare, line 4: Same here

agrare (changes requested):

spec/models/manageiq/providers/embedded_terraform/automation_manager/configuration_script_source_spec.rb
agrare: E.g. you can see it is causing issues just while running the tests
```
** ManageIQ master, codename: Spassky
/home/grare/adam/src/manageiq/manageiq-providers-embedded_terraform/spec/models/manageiq/providers/embedded_terraform/automation_manager/configuration_script_source_spec.rb:1: warning: already initialized constant TERRAFORM_RUNNER_URL
/home/grare/adam/src/manageiq/manageiq-providers-embedded_terraform/spec/lib/terraform/runner_spec.rb:4: warning: previous definition of TERRAFORM_RUNNER_URL was here
```

agrare:
Re: `Revert "fix for codeclient warning - remove`
No need to commit a revert to this branch, you can just force push the prior commit

Fryguy:
Rekicking the specs after merging #76

putmanoj:
> Re: `Revert "fix for codeclient warning - remove` No need to commit a revert to this branch, you can just force push the prior commit

Sure, thanks, will do that

spec/models/manageiq/providers/embedded_terraform/automation_manager/configuration_script_source_spec.rb
putmanoj: Fixed

app/models/service_template_terraform_template.rb
```diff
@@ -17,11 +17,28 @@ def self.create_catalog_item(options, _auth_user)
 
     transaction do
       create_from_options(options).tap do |service_template|
+        dialog_ids = service_template.send(:create_dialogs, config_info)
```
agrare, line 20: Why use `send` here?  This method doesn't appear to be private and the method name isn't dynamic so I think you can just
```suggestion
        dialog_ids = service_template.create_dialogs(config_info)
```

app/models/service_template_terraform_template.rb
```diff
@@ -41,4 +58,39 @@ def terraform_template(action)
...
+
+  def create_new_dialog(dialog_name, terraform_template, extra_vars)
+    Dialog::TerraformTemplateServiceDialog.create_dialog(dialog_name, terraform_template, extra_vars)
+  end
+
+  def validate_update_config_info(options)
+    opts = super
+    return unless options.key?(:config_info)
+
+    self.class.send(:validate_config_info, opts)
```
agrare, line 94: If you need to call this method from an instance variable I'd prefer you make it not a private_class_method and not use send

lib/terraform/runner.rb
```diff
@@ -180,6 +188,26 @@ def encoded_zip_from_directory(template_path)
         end
       end
 
+      # Parse Variables in Terraform Template
+      def template_variables(
+        template_path
+      )
```
agrare, line 194: This isn't a very long argument list :) can we one-line this?
```suggestion
      def template_variables(template_path)
```

spec/models/manageiq/providers/embedded_terraform/automation_manager/configuration_script_source_spec.rb
```diff
@@ -15,7 +17,25 @@
     let(:repos)              { Dir.glob(File.join(repo_dir, "*")) }
     let(:repo_dir_structure) { %w[hello_world.tf] }
 
+    let(:hello_world_vars_response) do
+      JSON.parse(File.read(File.join(__dir__, "../../../../../lib/terraform/runner/data/responses/hello-world-variables-success.json")))
```
agrare, line 21: You can also use the engine.root to get a path relative to the top of the plugin
```suggestion
      JSON.parse(ManageIQ::Providers::EmbeddedTerraform::Engine.root.join("spec/lib/terraform/runner/data/responses/hello-world-variables-success.json").read)
```

agrare (approved):
A few comments mostly about code style so please follow-up with another PR to address these, but nothing blocking functionality so I'm going to merge to get this going

app/models/service_template_terraform_template.rb
putmanoj: This came from ansible  implementation,  myself  was puzzled  why  'send' ?   
simply did not fix what was not broken

agrare: :+1: That's because their create_dialogs instance method is private, that isn't a good reason to use send but that is why it is needed there and not here.

putmanoj:
@miq-bot add_labels release-4.7/yes?, radjabov/yes

putmanoj:
> @putmanoj Cannot apply the following label because they are not recognized:
> 
> * `release-4.7/yes?`
> 
> All labels for `ManageIQ/manageiq-providers-embedded_terraform`: https://github.com/ManageIQ/manageiq-providers-embedded_terraform/labels

ok

Fryguy:
Backported to `radjabov` in commit 777f0d23f6ce94f6eff261958f4183562fc620c0.

```text
commit 777f0d23f6ce94f6eff261958f4183562fc620c0
Date:   Tue Oct 1 09:01:24 2024 -0400

    Merge pull request #75 from putmanoj/create-dialog-for-parsed-template-input-vars
    
    Parse terraform template & create new service dialog with parsed input vars
    
    (cherry picked from commit 6fd1b368e06c09ae478250ea77d01c7d25bcfcc1)
```

PR #78, opened 2024-10-16
Add method fetch terraform stdout for UI

putmanoj:
Add methods in Stack, to fetch terraform job stdout from TerraformRunner `api/stack`

### DIFF
--- a/app/models/manageiq/providers/embedded_terraform/automation_manager/stack.rb
+++ b/app/models/manageiq/providers/embedded_terraform/automation_manager/stack.rb
@@ -54,4 +54,59 @@ class ManageIQ::Providers::EmbeddedTerraform::AutomationManager::Stack < ManageI
   def raw_status
     Status.new(miq_task)
   end
+
+  # Intend to be called by UI to display stdout. The stdout is stored in MiqTask#task_results or #message if error
+  # Since the task_results may contain a large block of data, it is desired to remove the task upon receiving the data
+  def raw_stdout_via_worker(userid, format = 'txt')
+    unless MiqRegion.my_region.role_active?("embedded_terraform")
+      msg = "Cannot get standard output of this terraform-template because the embedded terraform role is not enabled"
+      return MiqTask.create(
+        :name    => 'terraform_stdout',
+        :userid  => userid || 'system',
+        :state   => MiqTask::STATE_FINISHED,
+        :status  => MiqTask::STATUS_ERROR,
+        :message => msg
+      ).id
+    end
+
+    options = {:userid => userid || 'system', :action => 'terraform_stdout'}
+    queue_options = {
+      :class_name  => self.class,
+      :method_name => 'raw_stdout',
+      :instance_id => id,
+      :args        => [format],
+      :priority    => MiqQueue::HIGH_PRIORITY,
+      :role        => nil
+    }
+
+    MiqTask.generic_action_with_callback(options, queue_options)
+  end
+
+  def raw_stdout(format = 'txt')
+    case format
+    when "json" then raw_stdout_json
+    when "html" then raw_stdout_html
+    else             raw_stdout_txt
+    end
+  end
+
+  def raw_stdout_json
+    raw_stdout_txt.split("\n")
+  end
+
+  def raw_stdout_txt
+    job = Job.find_by(:miq_task_id => miq_task.id)
+    terraform_stack_id = job.options[:terraform_stack_id]
+
+    if terraform_stack_id
+      response = Terraform::Runner.fetch_result_by_stack_id(terraform_stack_id)
+      response.message
+    end
+  end
+
+  def raw_stdout_html
+    text = raw_stdout_txt
+    text = _("No output available") if text.blank?
+    TerminalToHtml.render(text)
+  end
 end
--- a/app/models/manageiq/providers/embedded_terraform/automation_manager/stack.rb
+++ b/app/models/manageiq/providers/embedded_terraform/automation_manager/stack.rb
@@ -55,8 +55,7 @@ class ManageIQ::Providers::EmbeddedTerraform::AutomationManager::Stack < ManageI
     Status.new(miq_task)
   end
 
-  # Intend to be called by UI to display stdout. The stdout is stored in MiqTask#task_results or #message if error
-  # Since the task_results may contain a large block of data, it is desired to remove the task upon receiving the data
+  # Intend to be called by UI to display stdout. The stdout is stored in TerraformRunner(api/stack#message)
   def raw_stdout_via_worker(userid, format = 'txt')
     unless MiqRegion.my_region.role_active?("embedded_terraform")
       msg = "Cannot get standard output of this terraform-template because the embedded terraform role is not enabled"


app/models/manageiq/providers/embedded_terraform/automation_manager/stack.rb
```diff
@@ -54,4 +54,58 @@ def refresh
...
+    else             raw_stdout_txt
+    end
+  end
+
+  def raw_stdout_json
+    raw_stdout_txt.split("\n")
+  end
+
+  def raw_stdout_txt
+    job = Job.find_by(:miq_task_id => miq_task.id)
```
agrare, line 97: `MiqTask` has a `job` association https://github.com/ManageIQ/manageiq/blob/master/app/models/miq_task.rb#L36 so you can simply do `job = miq_task.job`

app/models/manageiq/providers/embedded_terraform/automation_manager/stack.rb
```diff
@@ -54,4 +54,58 @@ def refresh
...
+    terraform_stack_id = job.options[:terraform_stack_id]
+
+    if terraform_stack_id
+      response = Terraform::Runner.fetch_result_by_stack_id(terraform_stack_id)
+      response.message
+    end
```
agrare, line 103: Minor, but if you have the entire rest of the method inside a conditional another way to do this is to return e.g.
```suggestion
    return if terraform_stack_id.nil?
    
    response = Terraform::Runner.fetch_result_by_stack_id(terraform_stack_id)
    response.message
```

app/models/manageiq/providers/embedded_terraform/automation_manager/stack.rb
```diff
@@ -54,4 +54,58 @@ def refresh
...
+  def raw_stdout(format = 'txt')
+    case format
+    when "json" then raw_stdout_json
+    when "html" then raw_stdout_html
+    else             raw_stdout_txt
+    end
+  end
+
+  def raw_stdout_json
+    raw_stdout_txt.split("\n")
```
Fryguy, line 93: This feels like a copy-paste from the ansible one where the raw_stdout_txt is actually json (we ask ansible to give us the raw output in json format, then we massage it later).  In your case, I think it's just regular text, so this method would not be correct.  I don't think we use the json anywhere, so you can either delete the json paths, or raise a NotImplementError here

Fryguy:
@putmanoj Can you add some specs for these methods?  You can probably copy a lot from core's spec/models/manageiq/providers/embedded_ansible/automation_manager/job_spec.rb around line 143+

app/models/manageiq/providers/embedded_terraform/automation_manager/stack.rb
putmanoj: text which is,  terraform stdout with terminal color codes.

yes, json format is not required,  was not sure, if it required in ManageIQ anywhere ...  
As suggested, will raise the raise NotImplementedError,

app/models/manageiq/providers/embedded_terraform/automation_manager/stack.rb
```diff
@@ -83,29 +83,35 @@ def raw_stdout_via_worker(userid, format = 'txt')
...
-    raw_stdout_txt.split("\n")
-  end
-
   def raw_stdout_txt
-    job = miq_task.job
-    terraform_stack_id = job.options[:terraform_stack_id]
+    data = terraform_runner_stack_data
 
-    return if terraform_stack_id.blank?
+    return '' if data.nil?
```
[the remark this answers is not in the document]
agrare: Or you could just do `data&.message` and skip the explicit return, up to you

app/models/manageiq/providers/embedded_terraform/automation_manager/stack.rb
putmanoj: Deleting the json path

spec/models/manageiq/providers/embedded_terraform/automation_manager/stack_spec.rb
```diff
@@ -30,4 +30,125 @@
...
+      require 'json'
+      let(:terraform_runner_url) { "https://1.2.3.4:7000" }
+      let(:hello_world_retrieve_response) do
+        JSON.parse(File.read(File.join(__dir__, "../../../../../lib/terraform/runner/data/responses/hello-world-retrieve-success.json")))
+      end
+      let(:miq_task) do
+        miq_task = FactoryBot.create(:miq_task)
+        miq_task.job = job
+        miq_task
+      end
```
agrare, line 49: You can do attribute assignment right in the `FactoryBot.create` call:
```suggestion
      let(:miq_task) { FactoryBot.create(:miq_task, :job => job) }
```

spec/models/manageiq/providers/embedded_terraform/automation_manager/stack_spec.rb
```diff
@@ -30,4 +30,125 @@
...
+          job.options = {
+            :terraform_stack_id => hello_world_retrieve_response['stack_id']
+          }
+        end
+      end
+
+      let(:terraform_runner_stdout) { hello_world_retrieve_response['message'] }
+      let(:terraform_runner_stdout_html) { TerminalToHtml.render(terraform_runner_stdout) }
+
+      retrieve_stub = nil
```
agrare, line 63: This doesn't look right...

spec/models/manageiq/providers/embedded_terraform/automation_manager/stack_spec.rb
```diff
@@ -30,4 +30,125 @@
...
+        stack.miq_task = miq_task
+
+        stub_const("ENV", ENV.to_h.merge("TERRAFORM_RUNNER_URL" => terraform_runner_url))
+
+        retrieve_stub = stub_request(:post, "#{terraform_runner_url}/api/stack/retrieve")
+                        .with(:body => hash_including({:stack_id => hello_world_retrieve_response['stack_id']}))
+                        .to_return(
+                          :status => 200,
+                          :body   => hello_world_retrieve_response.to_json
+                        )
```
agrare, line 75: Oh you are trying to get access to this output from outside the before block, I'm not sure we really care about the have_been_requested.times(1) if the output is correct that should be sufficient so you could just drop the `retrieve_stub =` part

spec/models/manageiq/providers/embedded_terraform/automation_manager/stack_spec.rb
```diff
@@ -30,4 +30,125 @@
...
+        end
+      end
+
+      let(:terraform_runner_stdout) { hello_world_retrieve_response['message'] }
+      let(:terraform_runner_stdout_html) { TerminalToHtml.render(terraform_runner_stdout) }
+
+      retrieve_stub = nil
+
+      before do
+        stack.miq_task = miq_task
```
agrare, line 66: It looks like stack is already initialized with miq_task up on line 3

spec/models/manageiq/providers/embedded_terraform/automation_manager/stack_spec.rb
```diff
@@ -30,4 +30,125 @@
...
+
+      it "nil" do
+        expect(stack.raw_stdout).to eq ""
+      end
+    end
+
+    context "when miq_task.job is missing" do
+      before do
+        stack.miq_task = nil
+      end
```
agrare, line 128: I would do `let(:miq_task) { nil }` or override the stack definition like `let(:stack) { FactoryBot.create(:terraform_stack) }` rather than do this in a before block

spec/models/manageiq/providers/embedded_terraform/automation_manager/stack_spec.rb
```diff
@@ -30,4 +30,125 @@
...
+
+      it_behaves_like "terraform runner stdout not available from miq_task"
+    end
+
+    context "when miq_task present, but missing miq_task.job" do
+      before do
+        stack.miq_task = FactoryBot.create(:miq_task)
+        stack.miq_task.job = nil
+      end
```
agrare, line 137: Similarly it would be better to override the miq_task definition to not include the job like `let(:miq_task) { FactoryBot.create(:miq_task, :job => nil) }`

spec/models/manageiq/providers/embedded_terraform/automation_manager/stack_spec.rb
```diff
@@ -30,4 +30,125 @@
...
+      it_behaves_like "terraform runner stdout not available from miq_task"
+    end
+
+    context "when miq_task.job.options present but missing terraform_stack_id" do
+      before do
+        stack.miq_task = FactoryBot.create(:miq_task)
+        stack.miq_task.job = ManageIQ::Providers::EmbeddedTerraform::AutomationManager::Job.create_job(template, {}, {}, []).tap do |job|
+          job.state = "waiting_to_start"
+          job.options = {}
+        end
```
agrare, line 148: Think you know what I'm going to say :laughing: better to override the job definition only and let the other let() calls work normally

spec/models/manageiq/providers/embedded_terraform/automation_manager/stack_spec.rb
putmanoj: both are in different 'describe' block

spec/models/manageiq/providers/embedded_terraform/automation_manager/stack_spec.rb
putmanoj: I guess, it is more readable this way,  
when looking at, what the test is doing

agrare: `let(:miq_task) { nil }` is pretty clear what is happening, and you can get some weird behavior with before if you're not careful since they all run at each level it is easier to get messed up

putmanoj: Do you mean in this way ?

```
    context "when miq_task is missing" do
      let(:miq_task) { nil }

      it_behaves_like "terraform runner stdout not available from miq_task"
    end

    context "when miq_task present, but missing miq_task.job" do
      let(:miq_task) { FactoryBot.create(:miq_task, :job => nil) }

      it_behaves_like "terraform runner stdout not available from miq_task"
    end

    context "when miq_task.job.options present but missing terraform_stack_id" do
      let(:job) do
        ManageIQ::Providers::EmbeddedTerraform::AutomationManager::Job.create_job(template, {}, {}, []).tap do |job|
          job.state = "waiting_to_start"
          job.options = {}
        end
      end

      let(:miq_task) { FactoryBot.create(:miq_task, :job => job) }

      it_behaves_like "terraform runner stdout not available from miq_task"
    end
```

if yes,  then 2nd & 3rd context, did not work as expected ...

putmanoj:
> @putmanoj Can you add some specs for these methods? You can probably copy a lot from core's spec/models/manageiq/providers/embedded_ansible/automation_manager/job_spec.rb around line 143+

specs added

spec/models/manageiq/providers/embedded_terraform/automation_manager/stack_spec.rb
```diff
@@ -30,4 +30,161 @@
...
+        JSON.parse(File.read(File.join(__dir__, "../../../../../lib/terraform/runner/data/responses/hello-world-retrieve-success.json")))
+      end
+      let(:miq_task) { FactoryBot.create(:miq_task, :job => job) }
+
+      let(:job) do
+        ManageIQ::Providers::EmbeddedTerraform::AutomationManager::Job.create_job(template, {}, {}, []).tap do |job|
+          job.state = "finished"
+          job.options = {
+            :terraform_stack_id => hello_world_retrieve_response['stack_id']
+          }
```
agrare, line 51: I think a FactoryBot factory would be a better way to set up these Job objects

spec/models/manageiq/providers/embedded_terraform/automation_manager/stack_spec.rb
```diff
@@ -30,4 +30,161 @@
...
+
+  describe "#raw_stdout" do
+    let(:stack) { FactoryBot.create(:terraform_stack, :miq_task => miq_task) }
+    let(:template) { FactoryBot.create(:terraform_template) }
+
+    context "when miq_task.job present" do
+      let(:terraform_runner_url) { "https://1.2.3.4:7000" }
+      let(:hello_world_retrieve_response) do
+        require 'json'
+        JSON.parse(File.read(File.join(__dir__, "../../../../../lib/terraform/runner/data/responses/hello-world-retrieve-success.json")))
```
agrare, line 42: We might want to consider moving this up to `spec/data` or something if it is going to be used by more than just the `Terraform::Runner` specs since this `../../../etc...` makes it hard to read where this file lives

agrare (approved):
A number of items that we can clean up but overall looks good to me

spec/models/manageiq/providers/embedded_terraform/automation_manager/stack_spec.rb
agrare: https://github.com/ManageIQ/manageiq-providers-embedded_terraform/pull/80

spec/models/manageiq/providers/embedded_terraform/automation_manager/stack_spec.rb
putmanoj: Yes agree makes it harder to read,  
could make it like 
"../../../../../../spec/lib/terraform/runner/data/responses/hello-world-retrieve-success.json"

also maybe should move this common test data,  to somewhere like `<base-dir>/spec/data`.  then still look like
"../../../../../../spec/data/responses/hello-world-retrieve-success.json"

putmanoj: don't know if there more easier way to refer to test data resources without using "../../../../../../"

agrare: We can use the `ManageIQ::Providers::EmbeddedTerraform::Engine.root` Pathname to build a path from the root of the plugin

putmanoj: ok

Fryguy:
Backported to `radjabov` in commit ecb9a6cdac4edea31897a097fce91181bf18a86e.

```text
commit ecb9a6cdac4edea31897a097fce91181bf18a86e
Date:   Mon Oct 21 09:20:36 2024 -0400

    Merge pull request #78 from putmanoj/add-method-fetch-terraform-stdout-for-ui
    
    Add methods to fetch terraform stdout for showing in UI.
    
    (cherry picked from commit 4c66529cb5827112f4c7640f11b705e9258669cd)
```

PR #83, opened 2024-10-29
Add instructions on running opentofu-runner locally

agrare:
If you need to run the opentofu-runner locally for development without a full podified / systemd runtime you can run just the docker container directly.

This is now needed now in order to sync a terraform repo in order to parse variables.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,34 @@ See the section on plugins in the [ManageIQ Developer Setup](http://manageiq.org
 
 For quick local setup run `bin/setup`, which will clone the core ManageIQ repository under the *spec* directory and setup necessary config files. If you have already cloned it, you can run `bin/update` to bring the core ManageIQ code up to date.
 
+### Running Opentofu Runner locally
+
+First ensure that you have the opentofu-runner image pulled locally:
+```
+docker pull docker.io/manageiq/opentofu-runner:latest
+```
+
+If your database connection requires a password then create a secret:
+```
+echo '{"DATABASE_PASSWORD":"mypassword"}' | docker secret create opentofu-runner-secret -
+```
+
+And add `--secret opentofu-runner-secret` to the command below
+
+Now you can start your opentofu-runner:
+```
+docker run --name=opentofu-runner --rm --network=host --env NODE_ENV=development --env DATABASE_HOSTNAME=localhost --env DATABASE_NAME=vmdb_development --env DATABASE_USERNAME=root --env MEMCACHE_SERVERS=127.0.0.1:11211 --env PORT=6000 --expose=6000 docker.io/manageiq/opentofu-runner:latest
+```
+
+Verify that everything is working by checking `Terraform::Runner.available?` with a rails console:
+```
+$ TERRAFORM_RUNNER_URL=http://localhost:6000 rails c
+>> Terraform::Runner.available?
+=> true
+```
+
+To stop the opentofu-runner `docker stop opentofu-runner` in another terminal.
+
 ## License
 
 The gem is available as open source under the terms of the [Apache License 2.0](http://www.apache.org/licenses/LICENSE-2.0).


agrare:
NOTE I want to get this to the point where you can just `run_single_worker OpentofuWorker` this thing but we're not there yet.  The non-rails non-podified non-systemd process worker option right now expects `workers/opentofu_worker/worker` like the vmware non-rails event catcher.

README.md
```diff
@@ -14,6 +14,40 @@ See the section on plugins in the [ManageIQ Developer Setup](http://manageiq.org
 
 For quick local setup run `bin/setup`, which will clone the core ManageIQ repository under the *spec* directory and setup necessary config files. If you have already cloned it, you can run `bin/update` to bring the core ManageIQ code up to date.
 
+### Running Opentofu Runner locally
+First create the certs
+```
+mkdir certs
+openssl req -x509 -newkey rsa -days 1095 -keyout certs/tf_runner.key -out certs/tf_runner.crt -subj "/CN=server" -nodes -batch
+chmod 666 certs/tf_runner.crt certs/tf_runner.key
```
jrafanie, line 22: Is 666 the correct permissions?  I'm surprised something doesn't complain about the permissions being too permissive.  It's dev setup but still it's curious.

jrafanie (commented):
Nice, just minor nits and a question

README.md
agrare: Yeah I didn't like this but haven't dug into why, but with 0640 opentofu-runner fails with a permissions error

README.md
```diff
@@ -14,6 +14,40 @@ See the section on plugins in the [ManageIQ Developer Setup](http://manageiq.org
 
 For quick local setup run `bin/setup`, which will clone the core ManageIQ repository under the *spec* directory and setup necessary config files. If you have already cloned it, you can run `bin/update` to bring the core ManageIQ code up to date.
 
+### Running Opentofu Runner locally
+First create the certs
```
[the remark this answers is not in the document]
agrare: :+1: good catch, fixed

README.md
agrare: Okay `0644 ` works also which matches what we do on appliances, the issue is they run as a "stack" user in the container so we can't just do 0600 like I'd like to

jrafanie: What about 644?  Hopefully, if it's considered "others" it only needs to read the file.

jrafanie: lol, I didn't refresh and you already did it 🤣

README.md
```diff
@@ -14,6 +14,41 @@ See the section on plugins in the [ManageIQ Developer Setup](http://manageiq.org
...
+If your database connection requires a password then create a secret:
+```
+echo '{"DATABASE_PASSWORD":"mypassword"}' | docker secret create opentofu-runner-secret -
+```
+
+And add `--secret opentofu-runner-secret` to the command below
+
+Now you can start your opentofu-runner:
+```
+docker run --name=opentofu-runner --rm --network=host -v ./certs/tf_runner.crt:/opt/app-root/src/config/cert/tls.crt:z -v ./certs/tf_runner.key:/opt/app-root/src/config/cert/tls.key --env DATABASE_HOSTNAME=localhost --env DATABASE_NAME=vmdb_development --env DATABASE_USERNAME=root --env MEMCACHE_SERVERS=127.0.0.1:11211 --env PORT=6000 --expose=6000 docker.io/manageiq/opentofu-runner:latest
```
Fryguy, line 40: Is `--expose=6000` the same as `-p 6000:6000` ?

README.md
agrare: @jrafanie alternatively if I pass `--env NODE_ENV=development` I'm able to run this without SSL which would simplify things (no need to create certs), what do you think?

jrafanie: haha, it's so much simpler now 👍

jrafanie (approved):
LGTM

PR #86, opened 2024-11-08
Use terraform-runner /live api in Terraform::Runner.available? method

putmanoj:
This is a small change. 
* We have added the Loopback4 Health checks component https://loopback.io/doc/en/lb4/Health.html
* So now we use the `/live` api added by the component, in `Terraform::Runner.available?` method
* `/ping` api also will be available.

### DIFF
--- a/spec/lib/terraform/runner_spec.rb
+++ b/spec/lib/terraform/runner_spec.rb
@@ -22,8 +22,8 @@ RSpec.describe(Terraform::Runner) do
 
   describe "is .available" do
     before do
-      stub_request(:get, "#{terraform_runner_url}/ping")
-        .to_return(:status => 200, :body => {'count' => 0}.to_json)
+      stub_request(:get, "#{terraform_runner_url}/live")
+        .to_return(:status => 200, :body => {:status => "UP", :checks => []}.to_json)
     end
 
     it "check if terraform-runner service is available" do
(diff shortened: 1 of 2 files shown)


spec/lib/terraform/runner_spec.rb
```diff
@@ -20,8 +20,10 @@
 
   describe "is .available" do
     before do
-      stub_request(:get, "#{terraform_runner_url}/ping")
-        .to_return(:status => 200, :body => {'count' => 0}.to_json)
+      ENV["TERRAFORM_RUNNER_URL"] = "https://1.2.3.4:7000"
```
agrare, line 23: Is this needed since we do `stub_const("ENV", ENV.to_h.merge("TERRAFORM_RUNNER_URL" => terraform_runner_url))` up on line 11?

agrare:
@putmanoj can you squash these 4 commits into 1?  I don't normally push for this but this change is so simple I don't think these 4 commits really show any meaningful progress to be preserved in the git history.

agrare:
Also can you add a comment in the PR and commit description about why the change the /live vs /ping fixes the issue?

agrare:
cc @putmanoj https://github.com/ManageIQ/manageiq-providers-embedded_terraform/pull/86#issuecomment-2465292086 and https://github.com/ManageIQ/manageiq-providers-embedded_terraform/pull/86#issuecomment-2465293154 are still outstanding

putmanoj:
Done @agrare

agrare (approved):

agrare:
I'm not going to mark this for backport yet until we confirm that a version of the opentofu-runner will be in the target branch

PR #89, opened 2024-12-05
Add dialog fields based on terraform type constraints

putmanoj:
This is a enhancement to dynamic dialog field generation for terraform. 
This change provides better support for terraform input variables dialog generation, in regards terraform input variable type constraints, such as List,Map,Number,Boolean,nullable values 

* Instead of only Textbox field for input-vars with List & Map type constraint, 
   * now TextAreabox fields will be added.
   * The default JSON values are prettified in textarea fields,  for easier editing of the JSON values. 
   * Also the List & Map will have some limited validation support with fields using regex
* This also fixes earlier implementation, where for type such as List & Map, the value was sent as JSON string, 
   * now the JSON is converted to JSON object or when required to nil value, before sending to terraform-runner api's 
* Number type field a little better support with regex validation.
* Boolean type field values,
   * now Checkbox field is added 
   * user value is convert to true or false, before sending to terraform-runner api.

### DIFF
--- a/app/models/manageiq/providers/embedded_terraform/automation_manager/job.rb
+++ b/app/models/manageiq/providers/embedded_terraform/automation_manager/job.rb
@@ -31,16 +31,18 @@ class ManageIQ::Providers::EmbeddedTerraform::AutomationManager::Job < Job
       Terraform::Runner.delete_stack(
         options[:terraform_stack_id],
         template_path,
-        :input_vars  => decrypt_vars(input_vars),
-        :credentials => credentials,
-        :env_vars    => options[:env_vars]
+        :input_vars                  => decrypt_vars(input_vars),
+        :input_vars_type_constraints => input_vars_type_constraints,
+        :credentials                 => credentials,
+        :env_vars                    => options[:env_vars]
       )
     else
       response = Terraform::Runner.create_stack(
         template_path,
-        :input_vars  => decrypt_vars(input_vars),
-        :credentials => credentials,
-        :env_vars    => options[:env_vars]
+        :input_vars                  => decrypt_vars(input_vars),
+        :input_vars_type_constraints => input_vars_type_constraints,
+        :credentials                 => credentials,
+        :env_vars                    => options[:env_vars]
       )
       options[:terraform_stack_id] = response.stack_id
       save!
@@ -154,5 +156,14 @@ class ManageIQ::Providers::EmbeddedTerraform::AutomationManager::Job < Job
     FileUtils.rm_rf(options[:git_checkout_tempdir])
   rescue Errno::ENOENT
     nil
+  end
+
+  def input_vars_type_constraints
+    require 'json'
+    payload = JSON.parse(template.payload)
+    payload['input_vars'] || []
+  rescue => error
+    _log.error("Failure in parsing payload for template/#{template.id}, caused by #{error.message}")
+    []
   end
 end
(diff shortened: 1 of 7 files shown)


app/models/manageiq/providers/embedded_terraform/automation_manager/job.rb
```diff
@@ -155,4 +157,13 @@ def cleanup_git_repository
   rescue Errno::ENOENT
     nil
   end
+
+  def input_vars_type_constraints
```
agrare, line 161: This method name doesn't seem to match since it looks like it returns the whole input_vars key, would `template_input_vars` work maybe?

lib/terraform/runner/api_params.rb
```diff
@@ -42,6 +43,74 @@ def self.to_cam_parameters(vars)
...
+      # Normalize variables values, from ManageIQ values to Terraform Runner supported values
+      # @param input_vars       [Hash]  key/value pairs as input variables for the terraform-runner run job.
+      # @param type_constraints [Array] array of type constraints objects, from Terraform Runner
+      #
+      # @return [Array] Array of param objects [{name,value,secured}]
+      def self.to_normalized_cam_parameters(input_vars, type_constraints)
+        require 'json'
+
+        input_vars.map do |k, v|
+          type_constr = type_constraints.find { |e| e['name'] == k.to_s }
```
agrare, line 59: It would be more efficient to build a hash indexed on the name key once, maybe you can return this as a hash from the method above [[here]](https://github.com/ManageIQ/manageiq-providers-embedded_terraform/pull/89/files#diff-d9ae67c3672d928afae2d394a3f872f3c2179e5a8ee51c4ffdd1250d51760a5dR161)
```
(payload['input_vars'] || []).index_by { |v| v['name'] }
```

Then here you can just use `type_constraints[k]`

lib/terraform/runner/api_params.rb
```diff
@@ -42,6 +43,74 @@ def self.to_cam_parameters(vars)
...
+            begin
+              value = JSON.parse(value)
+            rescue JSON::ParserError
+              raise "The variable '#{key}' does not have valid #{expected_type.name} value"
+            end
+          end
+        end
+
+        if value.nil?
+          if is_required == true
```
agrare, line 105: The `== true` isn't necessary you can just do `if is_required`

app/models/manageiq/providers/embedded_terraform/automation_manager/job.rb
agrare: Actually I guess this is fine since you're using the template input_vars for type checking

lib/terraform/runner/api_params.rb
```diff
@@ -42,6 +43,74 @@ def self.to_cam_parameters(vars)
           to_cam_param(key, value)
         end
       end
+
+      require 'set'
+      TRUE_VALUES = ['T', 't', true, 'true', 'True', 'TRUE'].to_set
```
agrare, line 48: You were using `ActiveModel::Type::Boolean.new.cast` above any reason to not continue to use that here?

putmanoj: Yes, there is reason, we are using `TRUE_VALUES`.

And `ActiveModel::Type::Boolean`, uses `FALSE_VALUES`, https://api.rubyonrails.org/classes/ActiveModel/Type/Boolean.html .

It is more to do with ManageIQ Checkbox field,  though we configure it with boolean `true` or `false` when the field is selected, the result is string `'t'`, whereas when unchecked it blank string `''`.

Terraform considers blank string `''` as invalid for boolean type.
 So we can convert `''` to `nil`/`null`, which is expectable  for Terraform.

In fact `ActiveModel::Type::Boolean` converts `''` to  `nil`. 

But `''` or `nil`/`null` cannot be always same as `false`.

PR #93, opened 2025-03-21
Support reconfigure(update) action in Terraform::Runner client

putmanoj:
* This PRs adds support in Terraform::Runner client to apply changes in terraform input params values. ( to support ResourceAction::RECONFIGURE for Service instance )

* The code has been refactored to remove duplicate code, for running the TerraformRunner actions,  instead of multiple action methods,  we now have single `run` method with TerraformRunner actions enum-like-constants values.
   - removed methods like `create_stack`, `delete_stack`, etc instead we use `run` method


lib/terraform/runner/action_type.rb
```diff
@@ -4,14 +4,15 @@ class Runner
...
+      UPDATE = APPLY
       DELETE = 'api/stack/delete'.freeze
       CANCEL = 'api/stack/cancel'.freeze
       RETRIEVE = 'api/stack/retrieve'.freeze
       TEMPLATE_VARIABLES = 'api/template/variables'.freeze
 
       def self.action_endpoint(action_type)
         case action_type
-        when ActionType::CREATE, ActionType::APPLY, ActionType::DELETE,
+        when ActionType::CREATE, ActionType::UPDATE, ActionType::DELETE,
```
agrare, line 15: Just curious why replace APPLY with UPDATE here, but above create a duplicate of it?

lib/terraform/runner/request.rb
```diff
@@ -24,10 +24,13 @@ def initialize(action_type)
 
       def build_json_post_arguments
         case @action_type
-        when ActionType::CREATE, ActionType::APPLY, ActionType::DELETE
-          if !@request_arguments.key?(:cloud_providers)
-            @request_arguments[:cloud_providers] = []
-          end
+        when ActionType::CREATE
+          @request_arguments[:name] = random_stack_name if !@request_arguments.key?(:name)
```
agrare, line 28: Prefer `unless @request_arguments.key?(:name)` instead of `if !` for simple single arg conditionals like this

agrare: Unless `@request_arguments[:name] = nil` is a valid case you want to allow, you could simplify this as
```suggestion
          @request_arguments[:name] ||= random_stack_name
```

lib/terraform/runner/request.rb
```diff
@@ -24,10 +24,13 @@ def initialize(action_type)
...
+          @request_arguments[:cloud_providers] = [] if !@request_arguments.key?(:cloud_providers)
+          @request_arguments[:parameters] = [] if !@request_arguments.key?(:parameters)
```
agrare, line 30: Same with these you can do e.g.
```suggestion
          @request_arguments[:cloud_providers] ||= []
          @request_arguments[:parameters]      ||= []
```

lib/terraform/runner.rb
```diff
@@ -15,83 +15,53 @@ def available?
...
+              .name(options[:name])
+              .stack_id(options[:stack_id])
+          )
+
+          Terraform::Runner::ResponseAsync.new(response.stack_id)
+        when ActionType::TEMPLATE_VARIABLES
+          raise "Not supported action type in this method, instead use method parse_terraform_variables"
+        else
+          raise "Not supported action type '#{action_type}'"
+        end
```
agrare, line 64: I like to handle these failure cases early in the method versus having the entire thing indended under a case/conditional.  It reads a little better, getting all of the checks out of the way early and the rest of the method assumes you've made it past the gatekeeping.

You could e.g.:
```
raise "Not supported action type in this method, instead use method parse_terraform_variables" if action_type == ActionType::TEMPLATE_VARIABLES
raise "Not supported action type '#{action_type}'" unless ActionType.actions.include?(action_type)

response = run_terraform_runner_stack_api(
etc...
```

lib/terraform/runner/action_type.rb
putmanoj: UPDATE is like alias for APPLY,   was thinking perhaps UPDATE is more readable than APPLY, for Reconfiguration action.

Also CREATE essentially runs  `terraform apply` in the backend.

app/models/manageiq/providers/embedded_terraform/automation_manager/job.rb
```diff
@@ -26,24 +26,33 @@ def execute
...
-      Terraform::Runner.delete_stack(
-        options[:terraform_stack_id],
+      Terraform::Runner.run(
+        Terraform::Runner::ActionType::DELETE,
         template_path,
-        :input_vars                  => decrypt_vars(input_vars),
-        :input_vars_type_constraints => input_vars_type_constraints,
-        :credentials                 => credentials,
-        :env_vars                    => options[:env_vars]
+        runner_options.merge(:stack_id => options[:terraform_stack_id])
```
agrare, line 47: Not for this PR, but if we could get the action dynamically from ResourceAction this could be a single case

lib/terraform/runner.rb
```diff
@@ -15,99 +15,99 @@ def available?
...
+        response = run_terraform_runner_stack_api(
+          Request.new(action_type)
+            .template_path(template_path)
+            .credentials(options[:credentials])
+            .input_vars(options[:input_vars], options[:input_vars_type_constraints])
+            .tenant_id(stack_tenant_id)
+            .tags(options[:tags])
+            .env_vars(options[:env_vars])
+            .name(options[:name])
+            .stack_id(options[:stack_id])
```
agrare, line 57: Maybe it is just me but I had a hard time working out what was going on here :smile: 
Probably because these look like "getters" when they're really "setters", plus the return self and call the next method on the result of the previous.
Maybe in a follow-up this could be changed to something like
```
Request.new(action_type).tap do |req|
  req.template_path = template_path
  req.credentials = options[:credentials]
  ...
end
```
Or accept these as parameters in the `#initialize` method.

spec/lib/terraform/runner_spec.rb
```diff
@@ -198,19 +220,80 @@ def verify_req(req)
...
 
       it "is aliased as stop_stack" do
-        expect(Terraform::Runner.method(:stop_stack)).to(eq(Terraform::Runner.method(:stop_async)))
+        expect(Terraform::Runner.method(:stop)).to(eq(Terraform::Runner.method(:stop_async)))
       end
     end
 
-    describe '.delete_stack to run Retirement action' do
+    describe 'Update stack for Reconfiguration of created stack' do
+      update_stub = nil
```
agrare, line 228: Prefer using `let(:update_stub) { stub_request(:post, "https://1.2.3.4:7000/api/stack/apply") etc.... }` than setting a local variable at the `describe` context and then setting it in a `before {}`

spec/lib/terraform/runner_spec.rb
```diff
@@ -198,19 +220,80 @@ def verify_req(req)
...
+      retrieve_stub = nil
+
+      def verify_req(req)
+        body = JSON.parse(req.body)
+        expect(body["stack_id"]).to(eq(@hello_world_retrieve_update_response['stack_id']))
+        expect(body).to(have_key('templateZipFile'))
+        expect(body["parameters"]).to(eq([{"name" => "name", "value" => "Future-World", "secured" => "false"}]))
+        expect(body["cloud_providers"]).to(eq([]))
```
agrare, line 236:
```suggestion
        expect(body["stack_id"]).to eq(@hello_world_retrieve_update_response['stack_id'])
        expect(body).to have_key('templateZipFile')
        expect(body["parameters"]).to eq([{"name" => "name", "value" => "Future-World", "secured" => "false"}])
        expect(body["cloud_providers"]).to be_empty
```

spec/lib/terraform/runner_spec.rb
```diff
@@ -198,19 +220,80 @@ def verify_req(req)
...
+          {
+            :input_vars                  => input_vars,
+            :input_vars_type_constraints => input_vars_type_constraints,
+            :stack_id                    => @hello_world_retrieve_update_response['stack_id']
+          }
+        )
+        expect(update_stub).to(have_been_requested.times(1))
+
+        response = async_response.response
+        expect(retrieve_stub).to(have_been_requested.times(1))
```
agrare, line 278:
```suggestion
        expect(retrieve_stub).to have_been_requested.times(1)
```

spec/lib/terraform/runner_spec.rb
```diff
@@ -198,19 +220,80 @@ def verify_req(req)
...
+        expect(response.stack_id).to(eq(@hello_world_update_response['stack_id']))
+        expect(response.action).to(eq('APPLY'))
+        expect(response.stack_name).to(eq(@hello_world_update_response['stack_name']))
```
agrare, line 281:
```suggestion
        expect(response).to have_attributes(
          :stack_id   => @hello_world_update_response['stack_id'],
          :action     => "APPLY",
          :stack_name => @hello_world_update_response['stack_name']
        )
```

spec/lib/terraform/runner_spec.rb
```diff
@@ -198,19 +220,80 @@ def verify_req(req)
...
+
+        expect(response.status).to(eq('SUCCESS'), "terraform-runner failed with:\n#{response.status}")
```
agrare, line 283: I'm not sure this is testing the second return value properly, maybe
```suggestion
        expect(response.status).to eq(['SUCCESS', "terraform-runner failed with:\n#{response.status}"])
```

agrare: Also, is it normal to have the status of SUCCESS with a message of "terraform-runner failed with..."?

spec/lib/terraform/runner_spec.rb
```diff
@@ -198,19 +220,80 @@ def verify_req(req)
...
       delete_stub = nil
       delete_retrieve_stub = nil
 
       def verify_req(req)
         body = JSON.parse(req.body)
         expect(body["stack_id"]).to(eq(@hello_world_retrieve_delete_response['stack_id']))
         expect(body).to(have_key('templateZipFile'))
-        expect(body["parameters"]).to(eq([{"name" => "name", "value" => "New World", "secured" => "false"}]))
+        expect(body["parameters"]).to(eq([{"name" => "name", "value" => "Future-World", "secured" => "false"}]))
         expect(body["cloud_providers"]).to(eq([]))
```
agrare, line 297: Similar comment to above, the style we use for rspec expections is not to use parens around the `.to(eq())` so `.to eq(...)` is preferred

agrare:
Couple minor things in the specs but overall looking good

spec/lib/terraform/runner_spec.rb
putmanoj: the `let(:update_stub)`  lazily evaluated, so many not be available when http call is made. it will be evaluated,  when latter, expect expression uses it.

spec/lib/terraform/runner_spec.rb
putmanoj: it is not status value, it is 2nd argument for the expect(),   .., removing it, as not essential

spec/lib/terraform/runner_spec.rb
agrare: That is what `let!` is for if you need it evaluated before it is referenced.

app/models/manageiq/providers/embedded_terraform/automation_manager/job.rb
```diff
@@ -26,28 +26,32 @@ def execute
...
+      :input_vars_type_constraints => input_vars_type_constraints,
+      :credentials                 => credentials,
+      :env_vars                    => options[:env_vars]
+    }
+
+    # required for ResourceAction::RECONFIGURE, ResourceAction::RETIREMENT
+    unless action == ResourceAction::PROVISION
+      runner_options = runner_options.merge(
+        :stack_id => options[:terraform_stack_id]
       )
```
agrare, line 40: So two things, first if you just want to set a single key this is better:
```suggestion
      runner_options[:stack_id] = options[:terraform_stack_id]
      )
```

Second, if you did in the future want to merge a hash into an existing hash you can use the `!` version of merge:
`runner_options.merge!(:a => "a", :b => "b")` which is equivalent to `runner_options = runner_options.merge(:a => "a", :b => "b")`

app/models/manageiq/providers/embedded_terraform/automation_manager/job.rb
```diff
@@ -121,9 +125,14 @@ def template_relative_path
   end
 
   def stack_response
-    return if options[:terraform_stack_id].nil?
+    action                 = options[:action]
+    terraform_stack_id     = options[:terraform_stack_id]
+    terraform_stack_job_id = options[:terraform_stack_job_id]
+    $embedded_terraform_log.info("ResponseAsync for action:#{action}, stack_id:#{terraform_stack_id}, stack_job_id:#{terraform_stack_job_id}")
```
agrare, line 131: Thinking this should be `.debug()` since it just prints out the action/stack_id/job_id which isn't really helpful unless you're trying to e.g. debug how far through it is getting?

app/models/manageiq/providers/embedded_terraform/automation_manager/job.rb
```diff
@@ -165,7 +174,18 @@ def input_vars_type_constraints
...
   end
+
+  def terraform_runner_action_type(resource_action)
+    case resource_action
+    when ResourceAction::RECONFIGURE
+      Terraform::Runner::ActionType::UPDATE
+    when ResourceAction::RETIREMENT
+      Terraform::Runner::ActionType::DELETE
+    else
+      Terraform::Runner::ActionType::CREATE
```
agrare, line 188: Really minor, but if I pass in garbage it would translate to CREATE.  If you're trying for a 1:1 mapping I like an explicit last `when` and the `else` fallthrough is a `raise ArgumentError, "Invalid resource_action type #{resource_action}"`

app/models/manageiq/providers/embedded_terraform/automation_manager/job.rb
putmanoj: Yes, meant it to be debug ...,  missed this

app/models/manageiq/providers/embedded_terraform/automation_manager/job.rb
agrare: Won't hold this PR up for this since this is functional but can you fix this in a follow-up?

PR #94, opened 2025-04-08
Support reconfigure action for ServiceTerraformTemplate

putmanoj:
This PR has changes to support ResourceAction::RECONFIGURE action in embedded_terraform for Service instance.

Depends on (required):
https://github.com/ManageIQ/manageiq-content/pull/763

Related changes (optional)
https://github.com/ManageIQ/manageiq/pull/23405

Resolves https://jsw.ibm.com/browse/CP4AIOPS-5515


app/models/service_terraform_template.rb
```diff
@@ -70,15 +73,37 @@ def check_completed(action)
...
+    when ResourceAction::RECONFIGURE
+      # we need to reload, to pull updates to options from db,
+      # because the stack_id was saved in service instance in queue job
+      reload
+
+      job_options = options[job_option_key(action)]
+      orchestration_stack_id = job_options[:orchestration_stack_id]
+      $embedded_terraform_log.debug("find service_resource by resource_id:#{orchestration_stack_id}")
+
+      service_resources.find_by(:resource_id => orchestration_stack_id).try(:resource)
```
agrare, line 103: If you're going to query on polymorphic object you have to include the `_type` as well, e.g. if there are 3 service_resources `[OrchestrationStack(id: 2), OrchestrationStack(id: 10), Vm(id: 2)]` then this won't always return the orchestration stack
```suggestion
      service_resources.find_by(:resource_type => 'OrchestrationStack', :resource_id => orchestration_stack_id).try(:resource)
```

Adding the action in there probably wouldn't hurt, I do wonder if we could use the same for provision/retire below as well

putmanoj: Yes, we can use for provision/retire, as saving the orchestration_stack_id for all

app/models/manageiq/providers/embedded_terraform/automation_manager/job.rb
```diff
@@ -178,13 +177,20 @@ def input_vars_type_constraints
     {}
   end
 
+  def input_vars
+    extra_vars = options.dig(:input_vars, :extra_vars) || {}
+    input_vars = options.dig(:input_vars, :input_vars) || {}
+    # merge & over extra_vars with input_vars
+    extra_vars.deep_merge!(input_vars)
```
agrare, line 184: Are you trying to modify `options` or just return a new copy of the object?  If the latter you don't need `deep_merge!` you want just `deep_merge`

app/models/manageiq/providers/embedded_terraform/automation_manager/job.rb
```diff
@@ -178,13 +177,20 @@ def input_vars_type_constraints
...
+  end
+
   def terraform_runner_action_type(resource_action)
     case resource_action
     when ResourceAction::RECONFIGURE
       Terraform::Runner::ActionType::UPDATE
     when ResourceAction::RETIREMENT
       Terraform::Runner::ActionType::DELETE
-    when ResoureAction::PROVISION
+    when ResourceAction::PROVISION
```
agrare, line 193: Oops, missed this.  Is this covered by specs?

app/models/manageiq/providers/embedded_terraform/automation_manager/job.rb
```diff
@@ -178,13 +177,20 @@ def input_vars_type_constraints
     {}
   end
 
+  def input_vars
+    extra_vars = options.dig(:input_vars, :extra_vars) || {}
+    input_vars = options.dig(:input_vars, :input_vars) || {}
```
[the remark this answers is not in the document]
putmanoj: Yes, agree this looks weird to me also,, this from older code. 
 I have made note for myself to improve this,  wanted to do this, once  Reconfigure work is complete

app/models/manageiq/providers/embedded_terraform/automation_manager/job.rb
putmanoj: here we want to override key/values in `extra_vars`(key/value added by user in UI ) 
with key/values from `input_vars`(dynamic dialog generated for input-params from terraform configuration file ),  ultimately we want single input params hash for terraform-runner.

In-fact, we must deprecate this 'extra_vars' itself,  this came in, when we were referring the Ansible implementation ,  to create the embedded_terraform.  

Actually we don't need 'Variables & Default Values',  we only need dynamically generated 'Dialog'
Screenshot 2025-04-09 at 6 19 41 PM

app/models/manageiq/providers/embedded_terraform/automation_manager/job.rb
putmanoj: That is typo,  unfortunately not covered by specs.

agrare:
FYI you can add this to `.rubocop_local.yml` to resolve the `Do not introduce global variables.` warnings 
```
Style/GlobalVars:
  AllowedVariables:
  - $embedded_terraform_log
```

app/models/manageiq/providers/embedded_terraform/automation_manager/job.rb
agrare: > That is typo, unfortunately not covered by specs.

Okay we should add a spec to test this then.

app/models/service_terraform_template.rb
```diff
@@ -55,15 +73,34 @@ def check_completed(action)
...
+
     add_resource!(stack, :name => action)
+
+    $embedded_terraform_log.debug("Service(#{id}).launch_terraform_template(#{action}) ends")
   end
 
   def stack(action)
-    service_resources.find_by(:name => action, :resource_type => 'OrchestrationStack').try(:resource)
+    job_options = options[job_option_key(action)]
+    unless job_options && job_options.key?(:orchestration_stack_id)
```
agrare, line 94: I avoid using `unless` with anything more complicated than a single conditional.  Anything that tacks on `&&` or `||` and it takes up more brain compute cycles than is worth IMO.  Just use a normal `if job_options.nil? || !job_options.key?(:orchestration_stack_id)` or even better `if !job_options&.key?(:orchestration_stack_id)`

app/models/service_terraform_template.rb
```diff
@@ -55,15 +73,34 @@ def check_completed(action)
...
+      # we need to reload, to pull updates to options from db,
+      # because the stack_id was saved in service instance in a queue job
+      reload
```
agrare, line 97: Hm the reload here is odd, if this value is being set by someone else after we dequeue this item that feels extremely racy/timing dependent.  Where is this being set exactly?

app/models/service_terraform_template.rb
```diff
@@ -140,4 +200,36 @@ def translate_credentials!(options)
...
+    options[:dialog] = dialog_options
+    save!
+  end
+
+  def save_orchestration_stack_id(action, stack_id)
+    job_options = options[job_option_key(action)] || {}
+    job_options[:orchestration_stack_id] = stack_id
+
+    $embedded_terraform_log.info("save orchestration_stack_id:#{stack_id} with job_options for #{action}")
+    options[job_option_key(action)] = job_options
```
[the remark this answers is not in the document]
putmanoj: check my comments here https://github.com/ManageIQ/manageiq-providers-embedded_terraform/pull/94#discussion_r2037909932

app/models/service_terraform_template.rb
putmanoj: https://github.com/putmanoj/manageiq-providers-embedded_terraform/blob/support-reconfigure-action-with-service/app/models/service_terraform_template.rb#L36
If you see `queue_opts` has  service.id, not the actual object from the original context, so a new copy of service object is separatly fetched, somewhere in the queue job.

agrare: Okay if it is predictable enough I'd rather reload in the caller (e.g. after `save_orchestration_stack_id` perhaps?)

putmanoj: I feel may not work, as  the, `save_orchestration_stack_id` runs in the queue job context,  not the original context.

though, let me check as suggested, if it helps.

agrare: Yeah I might need to run this through locally, but "random" reloads to "fix" things is a pretty big red flag for instability

app/models/service_terraform_template.rb
putmanoj: done

putmanoj:
> FYI you can add this to `.rubocop_local.yml` to resolve the `Do not introduce global variables.` warnings
> 
> ```
> Style/GlobalVars:
>   AllowedVariables:
>   - $embedded_terraform_log
> ```

done

app/models/manageiq/providers/embedded_terraform/automation_manager/job.rb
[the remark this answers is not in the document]
putmanoj: @agrare 
Added a issue https://github.com/ManageIQ/manageiq-providers-embedded_terraform/issues/96 to add handle above change.

app/models/manageiq/providers/embedded_terraform/automation_manager/job.rb
putmanoj: Added a issue https://github.com/ManageIQ/manageiq-providers-embedded_terraform/issues/96 to address this change.

app/models/service_template_terraform_template.rb
```diff
@@ -40,21 +40,24 @@ def update_catalog_item(options, auth_user = nil)
...
     raise _("Must provide a configuration_script_payload_id") if info[:provision][:configuration_script_payload_id].nil?
 
     info
   end
 
-  private_class_method def self.retirement_default_info(prov_info)
+  private_class_method def self.retirement_or_reconfigure_default_info(prov_info)
     info = prov_info.deep_dup
     info.delete(:fqname)
     info
```
agrare, line 63: Can be in a follow-up PR but, this whole thing could be `prov_info.except(:fqname)` (in fact you might not even need a method in that case, the retirement_or_reconfigure_... name is a little weird)

app/models/service_terraform_template.rb
```diff
@@ -39,9 +41,25 @@ def execute(action)
...
     task    = MiqTask.wait_for_taskid(task_id)
+
+    $embedded_terraform_log.debug("Service(#{id}).execute(#{action}) ends, with task/#{task_id}:#{task}")
     raise task.message unless task.status_ok?
   end
 
+  def postprocess(action)
+    $embedded_terraform_log.debug("Service(#{id}).postprocess(#{action}) starts")
+    case action
+    when ResourceAction::RECONFIGURE
```
agrare, line 52: Why case/when for a single branch versus just `if action == ResourceAction::RECONFIGURE` ?

app/models/service_terraform_template.rb
```diff
@@ -123,15 +167,31 @@ def job_option_key(action)
...
+  def parse_dialog_options_only(action_options)
+    dialog_options = action_options[:dialog] || {}
+    params = {}
+    dialog_options.each do |attr, val|
+      if attr.start_with?("dialog_")
+        var_key = attr.sub(/^(password::)?dialog_/, '')
+        params[var_key] = val
+      end
+    end
+    params.blank? ? {} : {:input_vars => params}
```
agrare, line 183: Is this necessary?  I don't see how params could become `nil` which looks like the only time `params` would be `blank?=>true` but not `{}`.

app/models/service_terraform_template.rb
```diff
@@ -123,15 +167,31 @@ def job_option_key(action)
...
+  # - Service::service => ##
+  def parse_dialog_options_only(action_options)
+    dialog_options = action_options[:dialog] || {}
+    params = {}
+    dialog_options.each do |attr, val|
+      if attr.start_with?("dialog_")
+        var_key = attr.sub(/^(password::)?dialog_/, '')
+        params[var_key] = val
+      end
+    end
```
agrare, line 182: you can use `dialog_options.each_with_object do |(attr, val), params|` so you don't have to initialize a local var and then return it separately
```suggestion
    dialog_options.each_with_object({}) do |(attr, val), params|
      if attr.start_with?("dialog_")
        var_key = attr.sub(/^(password::)?dialog_/, '')
        params[var_key] = val
      end
    end
```

agrare (approved):

agrare:
@putmanoj I'm going to merge this but I don't want to lose track of the issue with `reload` and that `stack_id` since I believe there is a better way to handle that situation.  Can you open an issue to investigate that as well

agrare:
Hey @putmanoj any progress on the issues mentioned in https://github.com/ManageIQ/manageiq-providers-embedded_terraform/pull/94#issuecomment-2819014827 ?  This was merged with the understanding that those would be addressed right after.

putmanoj:
working on https://github.com/ManageIQ/manageiq-providers-embedded_terraform/issues/96, after this will address https://github.com/ManageIQ/manageiq-providers-embedded_terraform/pull/94#issuecomment-2819014827  ( via issue https://github.com/ManageIQ/manageiq-providers-embedded_terraform/issues/98 )

PR #95, opened 2025-04-14
Fix liveness & readiness probe endpoint for terraform-runner

putmanoj:
The liveness_probe & readiness_probe endpoint  for opentofu-runner are in-correct in the OpentofuWorker code

https://github.com/ManageIQ/manageiq-providers-embedded_terraform/blob/master/app/models/opentofu_worker.rb#L41
https://github.com/ManageIQ/manageiq-providers-embedded_terraform/blob/master/app/models/opentofu_worker.rb#L46

Instead of `/ready`,   ~~`/api/v1/ready`~~  has been used.

So, in the logs of opentofu-runner, we see ->

```
[2025-04-11T05:38:04.667] [ERROR] auth.provider - /api/v1/ready does not have valid token 
```

because all api endpoints need valid token for authentication,
expect  `/live`, `/ready`, `/health` & `/ping`

### DIFF
--- a/app/models/opentofu_worker.rb
+++ b/app/models/opentofu_worker.rb
@@ -38,12 +38,12 @@ class OpentofuWorker < MiqWorker
   end
 
   def add_liveness_probe(container_definition)
-    container_definition[:livenessProbe] = container_definition[:livenessProbe].except(:exec).merge(:httpGet => {:path => "/api/v1/ready", :port => SERVICE_PORT, :scheme => "HTTPS"})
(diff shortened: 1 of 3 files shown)


agrare:
@putmanoj should we update the `Terraform::Runner.available?` API call to use the readiness API instead of the liveness API?  Think it makes sense that we're checking when it is ready to receive traffic vs pod needs to be restarted

agrare (approved):

agrare:
@putmanoj LGTM but you need to fix the failing test

Fryguy:
Backported to spassky via merge of master into spassky

PR #99, opened 2025-05-08
Expose ServiceTerraformTemplate#execute without wait

kbrock:
Moving the wait_on_task into workflows - don't want to block this here

This has been split into an async and a blocking version.

- Automate interface expects execute to be complete after the return.
- Workflows needs to run async, so it would rather implement the wait_for_task logic itself.

### DIFF
--- a/spec/models/service_terraform_template_spec.rb
+++ b/spec/models/service_terraform_template_spec.rb
@@ -326,6 +326,62 @@ RSpec.describe ServiceTerraformTemplate do
     end
   end
 
+  describe "#execute_async" do
+    [
+      ResourceAction::PROVISION,
+      ResourceAction::RECONFIGURE,
+      ResourceAction::RETIREMENT
+    ].each do |action|
+      describe "with #{action} action" do
+        let!(:service) do
+          FactoryBot.create(:service_terraform_template).tap do |s|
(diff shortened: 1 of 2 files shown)


agrare:
Just a dumb naming question otherwise can't wait to get rid of the synchronous execute :heart:

kbrock:
update:
- renamed `execute_no_wait` to `execute_async`

spec/models/service_terraform_template_spec.rb
```diff
@@ -326,6 +326,62 @@
...
+        let!(:service) do
+          FactoryBot.create(:service_terraform_template).tap do |s|
+            s.add_resource!(stack1, :name => ResourceAction::PROVISION)
+            case action
+            when ResourceAction::RECONFIGURE
+              s.add_resource!(stack2, :name => ResourceAction::RECONFIGURE)
+            when ResourceAction::RETIREMENT
+              s.add_resource!(stack2, :name => ResourceAction::RECONFIGURE)
+              s.add_resource!(stack3, :name => ResourceAction::RETIREMENT)
+            end
```
agrare, line 345: Do we need these stacks to be added as resources for anything that is being tested here?

agrare (approved):
One question but nothing to hold this up

PR #100, opened 2025-05-23
Remove redundant extra_vars, as input_vars from service dialog is used for Terraform Template

### LINKED ISSUE #96: Remove redundant `Add Key-Value Pair` in UI / `:extra_vars` from input_vars for Terraform Template.
The `Variables & Default Values` > `Add Key-Value Pair` from **Service Catalog Item** in UI  is not required for Terraform Template.   We already have `Dialog`, when creates **Service Dialog** from input params for the Terraform Template.  

Image

The `Add Key-Value Pair` adds `:extra_vars` in `input_vars`,  don't need separate  `:extra_vars` for providing adding input params.  As this does not serve any additional purpose for Terraform Template.

So, we must remove the redundant `Add Key-Value Pair` in UI / `:extra_vars` from input_vars for Terraform Template.

putmanoj:
We remove redundant extra_vars, as the extra_vars is not (addiiontionally) needed as input params (unlike in ansible)

For input params, we now have the input params for Terraform Template coming from service dialog.  As the service dialog for the Terraform Template now can be dynamically generated from Terrraform Template input parameters in UI.

Also includes minor review comment changes from https://github.com/ManageIQ/manageiq-providers-embedded_terraform/pull/94

Fixes #96

### DIFF
--- a/app/models/manageiq/providers/embedded_terraform/automation_manager/job.rb
+++ b/app/models/manageiq/providers/embedded_terraform/automation_manager/job.rb
@@ -178,7 +178,7 @@ class ManageIQ::Providers::EmbeddedTerraform::AutomationManager::Job < Job
   end
 
   def input_vars
-    options.dig(:input_vars, :input_vars) || {}
+    ServiceTerraformTemplate.get_input_vars_from_job_options(options)
   end
 
   def terraform_runner_action_type(resource_action)
--- a/app/models/manageiq/providers/embedded_terraform/automation_manager/job.rb
+++ b/app/models/manageiq/providers/embedded_terraform/automation_manager/job.rb
@@ -178,10 +178,6 @@ class ManageIQ::Providers::EmbeddedTerraform::AutomationManager::Job < Job
   end
 
   def input_vars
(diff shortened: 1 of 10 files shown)


app/models/manageiq/providers/embedded_terraform/automation_manager/job.rb
```diff
@@ -178,10 +178,7 @@ def input_vars_type_constraints
   end
 
   def input_vars
-    extra_vars = options.dig(:input_vars, :extra_vars) || {}
-    input_vars = options.dig(:input_vars, :input_vars) || {}
-    # merge & over extra_vars with input_vars
-    extra_vars.deep_merge!(input_vars)
+    ServiceTerraformTemplate.get_input_vars_from_job_options(options)
```
agrare, line 181: Do we need to handle the old style in two different places?  Seems like the Service when ordered could format the input_vars properly so the Job can just take the new style

agrare: :+1: awesome

agrare (approved):

PR #102, opened 2025-06-03
Change podman root containers/storage directory

agrare:
https://github.com/ManageIQ/manageiq/issues/23470

### DIFF
--- a/app/models/opentofu_worker.rb
+++ b/app/models/opentofu_worker.rb
@@ -10,7 +10,8 @@ class OpentofuWorker < MiqWorker
     %i[workers worker_base opentofu_worker]
   ]
 
-  OPENTOFU_RUNTIME_DIR = "/var/lib/manageiq/opentofu-runner".freeze
+  MIQ_CONTAINERS_ROOT_DIR = "/var/lib/manageiq/containers/storage".freeze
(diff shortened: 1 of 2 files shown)


app/models/opentofu_worker.rb
```diff
@@ -10,7 +10,8 @@ class OpentofuWorker < MiqWorker
     %i[workers worker_base opentofu_worker]
   ]
 
-  OPENTOFU_RUNTIME_DIR = "/var/lib/manageiq/opentofu-runner".freeze
+  MIQ_CONTAINERS_ROOT_DIR = "/var/lib/manageiq/containers/storage".freeze
```
Fryguy, line 13: How will this affect local dev env (and I guess test env)?  I'm wondering if we should gate this on Rails.env.production?

agrare: https://github.com/ManageIQ/manageiq-providers-embedded_terraform?tab=readme-ov-file#running-opentofu-runner-locally is how you run this worker locally.

Because this doesn't go through run_single_worker this has a different way to run it for development at least at present.

Fryguy: Ok yeah good point.

PR #106, opened 2025-07-01
Drop VCR secrets since we don't use VCR in this repo

Fryguy:
Depends on:
- [x] https://github.com/ManageIQ/manageiq/pull/23497

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,6 @@
 /.yardoc
 /Gemfile.lock
 /_yardoc/
-/config/secrets.yml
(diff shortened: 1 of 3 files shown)


Fryguy:
@agrare Do we even use secrets here?  Should I just drop these secrets?

agrare:
@Fryguy we don't use VCR or Secrets so yeah we can just drop them completely

Fryguy:
Ok let me change this PR then

agrare (approved):

PR #115, opened 2026-02-13
Add workflow provision & state machine 

putmanoj:
# Changes in the PR
- Add deprecation label for `generic_terraform_template`
- Add new catalog type `embedded_terraform` for workflow provision
- Will show templates in Configuration Script tab in UI
- Uses ConfigurationScript for ServiceTemplate workflow provision

### DIFF
--- a/app/models/manageiq/providers/embedded_terraform/automation_manager/configuration_script.rb
+++ b/app/models/manageiq/providers/embedded_terraform/automation_manager/configuration_script.rb
@@ -1,5 +1,8 @@
 class ManageIQ::Providers::EmbeddedTerraform::AutomationManager::ConfigurationScript < ManageIQ::Providers::EmbeddedAutomationManager::ConfigurationScript
-  has_many :stacks, :class_name => "ManageIQ::Providers::EmbeddedTerraform::AutomationManager::Stack", :inverse_of => :configuration_script, :dependent => :nullify
+  alias terraform_template parent
+
+  delegate :stacks, :to => :terraform_template
+  delegate :run,    :to => :terraform_template
 
   def self.manager_class
     module_parent
@@ -7,16 +10,5 @@ class ManageIQ::Providers::EmbeddedTerraform::AutomationManager::ConfigurationSc
 
   def my_zone
     manager&.my_zone
-  end
-
-  def run(vars = {}, _userid = nil)
-    env_vars    = vars.delete(:env) || {}
-    credentials = vars.delete(:credentials)
-    action = vars.delete(:action) || ResourceAction::PROVISION
-    terraform_stack_id = vars.delete(:terraform_stack_id)
-
-    self.class.module_parent::Job.create_job(
-      self, env_vars, vars, credentials, :action => action, :terraform_stack_id => terraform_stack_id
-    ).tap(&:signal_start)
   end
 end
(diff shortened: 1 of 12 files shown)


Fryguy:
@putmanoj Please edit the commit message (no need to say "initial implementation"...just say what it does) because it's causing the commit message to get truncated.

putmanoj:
> Does this depend on #116 at all?

Not dependant on #116,  though it has changes from @agrare from his earlier branch (workflow)

agrare:
@putmanoj can you rebase on master now that #119 is merged?

app/models/manageiq/providers/embedded_terraform/automation_manager/template.rb
```diff
@@ -1,3 +1,23 @@
...
+  end
+
+  def configuration_script
+    @configuration_script ||= self.class.module_parent::ConfigurationScript.find_or_create_by(
+      :parent_id                      => id,
+      :configuration_script_source_id => configuration_script_source_id,
+      :name                           => name,
+      :manager_id                     => manager_id
+    )
+  end
```
agrare, line 17: We shouldn't be creating records on access.  You should be able to use `children.first` since we have a `1:1` relationship of `scripts<->payloads`

app/models/manageiq/providers/embedded_terraform/automation_manager/configuration_script.rb
```diff
@@ -9,14 +10,7 @@ def my_zone
...
-    env_vars    = vars.delete(:env) || {}
-    credentials = vars.delete(:credentials)
-    action = vars.delete(:action) || ResourceAction::PROVISION
-    terraform_stack_id = vars.delete(:terraform_stack_id)
-
-    self.class.module_parent::Job.create_job(
-      self, env_vars, vars, credentials, :action => action, :terraform_stack_id => terraform_stack_id
-    ).tap(&:signal_start)
+  def terraform_template
+    parent
```
agrare: Prefer using `alias` instead of a method that just calls an AR association.

agrare:
This looks pretty close, just a couple of comments and a rebase needed.

app/models/manageiq/providers/embedded_terraform/automation_manager/template.rb
putmanoj: Though I agree, we should not create on access, but this was added for backward compatibility for records which did not have ConfigurationScript, and also for earlier change when :run method got removed from Template, moved to ConfigurationScript.   

Yes, now that :run method is back in Template, we don't need this.

putmanoj:
Rebased with latest from master

app/models/manageiq/providers/embedded_terraform/automation_manager/template.rb
putmanoj: We need it to maintain relation with Stack, as suggested added, 

```
  def configuration_script
    children&.first
  end
```

agrare: > but this was added for backward compatibility for records which did not have ConfigurationScript

I created a schema migration to take care of this case: https://github.com/ManageIQ/manageiq-schema/pull/837

agrare: Ah I was looking at an older diff where you were still calling .find_or_create_configuration_script, lgtm now

agrare (approved):

Fryguy:
🎉

Fryguy:
Backported to `tal` in commit 8de14c8ab3435c714f6fe2f26675ac1d1cf19420.

```text
commit 8de14c8ab3435c714f6fe2f26675ac1d1cf19420
Date:   Thu Mar 12 10:42:13 2026 -0400

    Merge pull request #115 from putmanoj/workflow
    
    Add workflow provision & state machine
    
    (cherry picked from commit b08ce87fb291ceb73ad05a90ad9ed7ba70ddb718)
```

PR #122, opened 2026-03-26
Add input vars and credential to ConfigurationScript provisioning & retirement

putmanoj:
Depends on:

- [x] https://github.com/ManageIQ/manageiq-ui-classic/pull/9949
- [x] https://github.com/ManageIQ/manageiq-ui-classic/pull/9950

### DIFF
--- a/app/models/manageiq/providers/embedded_terraform/automation_manager/job.rb
+++ b/app/models/manageiq/providers/embedded_terraform/automation_manager/job.rb
@@ -12,10 +12,8 @@ class ManageIQ::Providers::EmbeddedTerraform::AutomationManager::Job < Job
   end
 
   def target=(target)
-    return if target.nil?
-
-    self.target_class = target.class.name
-    self.target_id = target.id
+    self.target_class = target&.class&.name
+    self.target_id    = target&.id
   end
 
   def start
--- a/app/models/manageiq/providers/embedded_terraform/automation_manager/job.rb
+++ b/app/models/manageiq/providers/embedded_terraform/automation_manager/job.rb
@@ -11,6 +11,13 @@ class ManageIQ::Providers::EmbeddedTerraform::AutomationManager::Job < Job
     )
   end
 
+  def target=(target)
+    return if target.nil?
+
+    self.target_class = target.class.name
+    self.target_id = target.id
+  end
+
   def start
     queue_signal(:pre_execute)
   end
--- a/app/models/manageiq/providers/embedded_terraform/automation_manager/provision/state_machine.rb
+++ b/app/models/manageiq/providers/embedded_terraform/automation_manager/provision/state_machine.rb
@@ -4,11 +4,11 @@ module ManageIQ::Providers::EmbeddedTerraform::AutomationManager::Provision::Sta
   end
 
   def provision
-    stack_opts = service.stack_opts
-    stack = stack_klass.create_stack(source.terraform_template, stack_opts)
+    stack_opts = service.stack_opts(ResourceAction::PROVISION, options)
+    stack = stack_klass.create_stack(source.terraform_template, stack_opts.dup)
 
     phase_context[:stack_id] = stack.id
-    connect_to_service!(stack, :name => "Provision")
+    connect_to_service!(stack, {:name => "Provision", :options => stack_opts})
 
     save!
 
--- a/app/models/manageiq/providers/embedded_terraform/automation_manager/provision/state_machine.rb
+++ b/app/models/manageiq/providers/embedded_terraform/automation_manager/provision/state_machine.rb
@@ -4,7 +4,8 @@ module ManageIQ::Providers::EmbeddedTerraform::AutomationManager::Provision::Sta
   end
 
   def provision
-    stack = stack_klass.create_stack(source.terraform_template)
+    stack_opts = service.stack_opts
+    stack = stack_klass.create_stack(source.terraform_template, stack_opts)
 
     phase_context[:stack_id] = stack.id
     connect_to_service!(stack, :name => "Provision")
--- a/app/models/manageiq/providers/embedded_terraform/automation_manager/provision/state_machine.rb
+++ b/app/models/manageiq/providers/embedded_terraform/automation_manager/provision/state_machine.rb
@@ -69,25 +69,22 @@ module ManageIQ::Providers::EmbeddedTerraform::AutomationManager::Provision::Sta
     return if stack.nil? || service.nil?
 
     job_options = stack.miq_task&.job&.options
-
     return if job_options.nil?
 
     terraform_runner_stack_id = job_options[:terraform_stack_id]
     terraform_runner_stack_job_id = job_options[:terraform_stack_job_id]
-
     return if terraform_runner_stack_id.blank?
 
     stack_resource = service.service_resources.find_by(:resource => stack)
-
     return if stack_resource.nil?
 
-    unless stack_resource.update(
+    stack_resource.update!(
       :options => stack_resource.options.merge(
         "terraform_runner_stack_id"     => terraform_runner_stack_id,
         "terraform_runner_stack_job_id" => terraform_runner_stack_job_id
       )
     )
-      $embedded_terraform_log.warn("Failed to update stack resource options")
-    end
+  rescue => err
+    $embedded_terraform_log.warn("Failed to update stack resource options : #{err.message}")
   end
 end
--- a/app/models/manageiq/providers/embedded_terraform/automation_manager/provision/state_machine.rb
+++ b/app/models/manageiq/providers/embedded_terraform/automation_manager/provision/state_machine.rb
@@ -85,6 +85,6 @@ module ManageIQ::Providers::EmbeddedTerraform::AutomationManager::Provision::Sta
       )
     )
   rescue => err
-    $embedded_terraform_log.warn("Failed to update stack resource options : #{err.message}")
+    $embedded_terraform_log.warn("Failed to update stack resource options: #{err.message}")
   end
 end
--- a/app/models/manageiq/providers/embedded_terraform/automation_manager/provision/state_machine.rb
+++ b/app/models/manageiq/providers/embedded_terraform/automation_manager/provision/state_machine.rb
@@ -24,6 +24,8 @@ module ManageIQ::Providers::EmbeddedTerraform::AutomationManager::Provision::Sta
   end
 
   def post_provision
+    update_stack_resource_data!
+
     if succeeded?
       signal :mark_as_completed
     else
@@ -54,5 +56,38 @@ module ManageIQ::Providers::EmbeddedTerraform::AutomationManager::Provision::Sta
 
   def stack
     @stack ||= stack_klass.find(phase_context[:stack_id])
+  end
+
+  private
+
+  # Updates the service resource with terraform runner stack information
+  # that will be used during retirement actions.
+  #
+  # @return [void]
+  # @note This method is called during post_provision phase
+  def update_stack_resource_data!
+    return if stack.nil? || service.nil?
+
+    job_options = stack.miq_task&.job&.options
+
+    return if job_options.nil?
+
+    terraform_runner_stack_id = job_options[:terraform_stack_id]
+    terraform_runner_stack_job_id = job_options[:terraform_stack_job_id]
+
+    return if terraform_runner_stack_id.blank?
+
+    stack_resource = service.service_resources.find_by(:resource => stack)
+
+    return if stack_resource.nil?
+
+    unless stack_resource.update(
+      :options => stack_resource.options.merge(
+        "terraform_runner_stack_id"     => terraform_runner_stack_id,
+        "terraform_runner_stack_job_id" => terraform_runner_stack_job_id
+      )
+    )
+      $embedded_terraform_log.warn("Failed to update stack resource options")
+    end
   end
 end
--- a/app/models/manageiq/providers/embedded_terraform/automation_manager/provision_workflow.rb
+++ b/app/models/manageiq/providers/embedded_terraform/automation_manager/provision_workflow.rb
@@ -9,4 +9,14 @@ class ManageIQ::Providers::EmbeddedTerraform::AutomationManager::ProvisionWorkfl
       build_ci_hash_struct(cs, %w[name description manager_name])
     end
   end
+
+  def allowed_credentials(*_args)
+    ManageIQ::Providers::EmbeddedTerraform::AutomationManager::TemplateCredential.all.map do |auth|
+      build_ci_hash_struct(auth, %w[name type])
+    end
+  end
+
+  def self.default_dialog_file
+    'miq_provision_configuration_script_embedded_terraform_dialogs'
+  end
 end
--- a/app/models/manageiq/providers/embedded_terraform/automation_manager/provision_workflow.rb
+++ b/app/models/manageiq/providers/embedded_terraform/automation_manager/provision_workflow.rb
@@ -11,8 +11,25 @@ class ManageIQ::Providers::EmbeddedTerraform::AutomationManager::ProvisionWorkfl
   end
 
   def allowed_credentials(*_args)
-    ManageIQ::Providers::EmbeddedTerraform::AutomationManager::TemplateCredential.all.map do |auth|
-      build_ci_hash_struct(auth, %w[name type])
+    credential_type          = get_value(values[:credential_type])
+    allowed_credential_klass = credential_type&.safe_constantize || self.class.module_parent::TemplateCredential
+    allowed_credential_klass.all.to_h do |auth|
+      [auth.id, auth.name]
+    end
+  end
+
+  def allowed_credential_types(*_args)
+    credential_id = get_value(values[:credential_id])
+    credential    = self.class.module_parent::TemplateCredential.find_by(:id => credential_id) if credential_id
+
+    credential_classes = credential.present? ? [credential.class] : ManageIQ::Providers::EmbeddedTerraform::AutomationManager::TemplateCredential.descendants
+    credential_classes.to_h do |klass|
+      next unless klass.const_defined?(:API_OPTIONS)
+
+      api_options = klass::API_OPTIONS
+      next unless api_options[:label]
+
+      [klass.name, api_options[:label]]
     end
   end
 
--- a/app/models/manageiq/providers/embedded_terraform/automation_manager/stack.rb
+++ b/app/models/manageiq/providers/embedded_terraform/automation_manager/stack.rb
@@ -9,6 +9,8 @@ class ManageIQ::Providers::EmbeddedTerraform::AutomationManager::Stack < ManageI
     alias raw_create_job raw_create_stack
 
     def create_stack(terraform_template, options = {})
+      $embedded_terraform_log.debug("Creating job from template(#{terraform_template.name}) with options: #{options}")
+
       authentications = collect_authentications(terraform_template.manager, options)
 
       job = raw_create_stack(terraform_template, options)
--- a/app/models/manageiq/providers/embedded_terraform/automation_manager/stack.rb
+++ b/app/models/manageiq/providers/embedded_terraform/automation_manager/stack.rb
@@ -53,7 +53,8 @@ class ManageIQ::Providers::EmbeddedTerraform::AutomationManager::Stack < ManageI
   end
 
   def retireable?
-    true
+    # if service is a ServiceEmbeddedTerraform, then it is not retireable, using raw_delete_stack
+    service.instance_of?(ServiceEmbeddedTerraform)
   end
 
   def raw_delete_stack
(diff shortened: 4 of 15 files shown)


app/models/manageiq/providers/embedded_terraform/automation_manager/provision/state_machine.rb
```diff
@@ -54,4 +55,8 @@ def stack_klass
   def stack
     @stack ||= stack_klass.find(phase_context[:stack_id])
   end
+
+  def service
+    @service ||= Service.find_by(:guid => options[:service_guid])&.becomes(ServiceEmbeddedTerraform)
```
agrare: Why the `becomes()` ?

app/models/manageiq/providers/embedded_terraform/automation_manager/provision_workflow.rb
```diff
@@ -1,4 +1,8 @@
 class ManageIQ::Providers::EmbeddedTerraform::AutomationManager::ProvisionWorkflow < ManageIQ::Providers::AutomationManager::ProvisionWorkflow
+  def self.base_model
+    ManageIQ::Providers::EmbeddedTerraform::AutomationManager::ProvisionWorkflow
+  end
+
```
agrare: Why is this needed?

app/models/manageiq/providers/embedded_terraform/automation_manager/provision/state_machine.rb
```diff
@@ -4,10 +4,11 @@ def run_provision
   end
 
   def provision
-    stack = stack_klass.create_stack(source.terraform_template)
+    stack_opts = service.stack_opts(ResourceAction::PROVISION, options)
+    stack = stack_klass.create_stack(source.terraform_template, stack_opts)
 
     phase_context[:stack_id] = stack.id
-    connect_to_service!(stack, :name => "Provision")
```
agrare, line 10: Why delete the provision name?  This is used in the `stack(action)` method below

app/models/manageiq/providers/embedded_terraform/automation_manager/provision/state_machine.rb
putmanoj: This was older commit, before your changes were pushed to master,  and after I rebased not required

agrare:
@putmanoj the old dialogs had a "cloud type" dropdown that filtered the credential types with a simple list.  This seems simpler than creating the new grid in the UI would that be an option for you here?

putmanoj:
@agrare  I do plan to improve the UI, to use dropdown, perhaps like in old dialog.  

For now prioritising completing the Retirement.  Will do the UI improvements  in another PR.

app/models/manageiq/providers/embedded_terraform/automation_manager/job.rb
```diff
@@ -193,4 +197,16 @@ def terraform_runner_action_type(resource_action)
       raise "Invalid resource_action type #{resource_action}"
     end
   end
+
+  def update_context(response)
+    ## Save in Job.context for miq_task, context info will be updated in miq_text.context_data
```
agrare:
```suggestion
    ## Save in Job.context for miq_task, context info will be updated in miq_task.context_data
```

app/models/manageiq/providers/embedded_terraform/automation_manager/stack.rb
```diff
@@ -48,6 +50,30 @@ def collect_authentications(manager, options)
     end
   end
 
+  def retireable?
+    true
+  end
+
+  def raw_delete_stack
+    $embedded_terraform_log.info("Delete Here to delete stack:#{id}")
```
agrare: Probably not necessary :D

app/models/manageiq/providers/embedded_terraform/automation_manager/stack.rb
```diff
@@ -48,6 +50,30 @@ def collect_authentications(manager, options)
...
+
+    terraform_template = configuration_script_payload
+    job_options = miq_task.context_data[:job_options]
+    job_options[:action] = ResourceAction::RETIREMENT
+    job_options[:terraform_stack_id] = miq_task.context_data[:terraform_runner_stack_id]
```
agrare: Should we merge the `terraform_runner_stack_id` in with the `job_options` in the provision above [[ref]](https://github.com/ManageIQ/manageiq-providers-embedded_terraform/pull/122/changes#diff-d9ae67c3672d928afae2d394a3f872f3c2179e5a8ee51c4ffdd1250d51760a5dR209) if we need it here anyway?

app/models/manageiq/providers/embedded_terraform/automation_manager/stack.rb
```diff
@@ -48,6 +50,30 @@ def collect_authentications(manager, options)
...
+
+    $embedded_terraform_log.debug("Run job to delete stack for template(#{terraform_template.name}) with options: #{job_options}")
+
+    delete_job = terraform_template.run(job_options)
+
+    $embedded_terraform_log.debug("Delete job created : #{delete_job.id}")
+
+    # raise MiqException::MiqOrchestrationProvisionError
```
agrare: I'm sure just leftover debugging

```suggestion
```

putmanoj: yes, will clean this up

app/models/manageiq/providers/embedded_terraform/automation_manager/stack.rb
putmanoj: The terraform_stack_id in job_options is for backward compatibility with  deprecated implementation in automate,  though thinking of doing something about that, in later commits

app/models/manageiq/providers/embedded_terraform/automation_manager/stack.rb
```diff
@@ -105,9 +138,20 @@ def raw_stdout_html
     TerminalToHtml.render(text)
   end
 
+  def service_resource
+    srs = ServiceResource.where(:resource => self)
+    @service_resource ||= srs&.first if !srs&.empty?
```
agrare: If it isn't possible to have >1 of these, then prefer `.find_by` over `.where.first`
If it is possible to have >1 of these then we need to improve the query to make sure it is distinct.

agrare: Also there is an association on `OrchestrationStack` for `service_resources` so you can just do `service_resources` instead of `ServiceResource.where(:resource => self)`

app/models/manageiq/providers/embedded_terraform/automation_manager/stack.rb
```diff
@@ -105,9 +138,20 @@ def raw_stdout_html
...
+  end
+
   private
 
   def terraform_runner_stack_data
+    sr_options = service_resource&.options || {}
+    terraform_runner_stack_id = sr_options["terraform_runner_stack_id"]
```
agrare:
```suggestion
    terraform_runner_stack_id = service_resource&.options&.dig("terraform_runner_stack_id")
```

app/models/manageiq/providers/embedded_terraform/automation_manager/provision/state_machine.rb
```diff
@@ -54,4 +57,25 @@ def stack_klass
   def stack
     @stack ||= stack_klass.find(phase_context[:stack_id])
   end
+
+  private
+
+  def reconnect_to_service!(stack, options)
+    service.remove_resource(stack)
+    connect_to_service!(stack, options)
```
agrare: Not a fan of remove_resource then add_resource, could we just find the service resource and update the options instead?

app/models/manageiq/providers/embedded_terraform/automation_manager/provision/state_machine.rb
```diff
@@ -54,4 +57,37 @@ def stack_klass
...
+  # that will be used during retirement actions.
+  #
+  # @return [void]
+  # @note This method is called during post_provision phase
+  def update_stack_resource_data!
+    return if stack.nil? || service.nil?
+
+    job_options = stack.miq_task&.job&.options
+
+    return if job_options.nil?
```
agrare, line 72:
```suggestion
    job_options = stack.miq_task&.job&.options
    return if job_options.nil?
```

app/models/manageiq/providers/embedded_terraform/automation_manager/provision/state_machine.rb
```diff
@@ -54,4 +57,37 @@ def stack_klass
...
+
+    terraform_runner_stack_id = job_options[:terraform_stack_id]
+    terraform_runner_stack_job_id = job_options[:terraform_stack_job_id]
+
+    return if terraform_runner_stack_id.blank?
```
agrare, line 76: Do we not need `terraform_runner_stack_job_id` to be set?

app/models/manageiq/providers/embedded_terraform/automation_manager/provision/state_machine.rb
```diff
@@ -54,4 +57,37 @@ def stack_klass
...
+
+    stack_resource = service.service_resources.find_by(:resource => stack)
+
+    return if stack_resource.nil?
```
agrare, line 79:
```suggestion
    stack_resource = service.service_resources.find_by(:resource => stack)
    return if stack_resource.nil?
```

app/models/manageiq/providers/embedded_terraform/automation_manager/provision/state_machine.rb
```diff
@@ -54,4 +57,37 @@ def stack_klass
...
+
+    unless stack_resource.update(
+      :options => stack_resource.options.merge(
+        "terraform_runner_stack_id"     => terraform_runner_stack_id,
+        "terraform_runner_stack_job_id" => terraform_runner_stack_job_id
+      )
+    )
+      $embedded_terraform_log.warn("Failed to update stack resource options")
+    end
```
agrare: The indentation is a little funky here,  maybe

```suggestion
    unless stack_resource.update(
        :options => stack_resource.options.merge(
          "terraform_runner_stack_id"     => terraform_runner_stack_id,
          "terraform_runner_stack_job_id" => terraform_runner_stack_job_id
        )
      )
      $embedded_terraform_log.warn("Failed to update stack resource options")
    end
```

Or


```suggestion
    stack_resource.update!(
      :options => stack_resource.options.merge(
        "terraform_runner_stack_id"     => terraform_runner_stack_id,
        "terraform_runner_stack_job_id" => terraform_runner_stack_job_id
      )
    )
  rescue
    $embedded_terraform_log.warn("Failed to update stack resource options")
```

app/models/manageiq/providers/embedded_terraform/automation_manager/stack/status.rb
```diff
@@ -1,11 +1,35 @@
 class ManageIQ::Providers::EmbeddedTerraform::AutomationManager::Stack::Status < OrchestrationStack::Status
-  attr_accessor :task_status
+  LIVE_STATUS_RUNNING = 'running'.freeze
+  LIVE_STATUS_CREATED = 'create_complete'.freeze
+  LIVE_STATUS_FAILED = 'failed'.freeze
+  LIVE_STATUS_DELETED = 'delete_complete'.freeze
```
agrare, line 5: Maybe just me but I find these easier to read if they are lined up

```suggestion
  LIVE_STATUS_RUNNING = 'running'.freeze
  LIVE_STATUS_CREATED = 'create_complete'.freeze
  LIVE_STATUS_FAILED  = 'failed'.freeze
  LIVE_STATUS_DELETED = 'delete_complete'.freeze
```

app/models/manageiq/providers/embedded_terraform/automation_manager/stack.rb
```diff
@@ -25,14 +27,16 @@ def create_stack(terraform_template, options = {})
         :miq_task                     => miq_task,
         :status                       => miq_task&.state,
         :start_time                   => miq_task&.started_on
-      )
+      ).tap do |stack|
+        job.target = stack
+        job.save!
+      end
```
agrare, line 32: ~~Why tap and save again instead of just adding `:target_class => stack.class.name, :target_id => stack.id` above in the `create!()` call?~~

Nevermind I got my stack/job mixed up

app/models/manageiq/providers/embedded_terraform/automation_manager/stack.rb
```diff
@@ -48,7 +52,54 @@ def collect_authentications(manager, options)
     end
   end
 
+  def retireable?
+    # if service is a ServiceEmbeddedTerraform, then it is not retireable, using raw_delete_stack
+    service.instance_of?(ServiceEmbeddedTerraform)
+  end
```
agrare, line 58: What does this do exactly?

app/models/manageiq/providers/embedded_terraform/automation_manager/stack.rb
```diff
@@ -48,7 +52,54 @@ def collect_authentications(manager, options)
...
+
   def refresh
+    # when retiring
+    return if retired? || error_retiring?
+
+    if retiring?
+      return if delete_miq_task.nil?
+
+      if raw_status.running? #  delete_job&.is_active?
+        delete_job&.poll_runner
```
agrare, line 96: Is delete_job actually an association here or you just set an instance variable?

agrare: Ah I see the method below

app/models/manageiq/providers/embedded_terraform/automation_manager/stack.rb
```diff
@@ -48,7 +52,54 @@ def collect_authentications(manager, options)
...
+      end
+      return
```
agrare: Usually reads a little better when you do `if ... else ... end` instead of returning from an if condition (unless it is error handling)

app/models/manageiq/providers/embedded_terraform/automation_manager/stack.rb
```diff
@@ -105,16 +164,53 @@ def raw_stdout_html
     TerminalToHtml.render(text)
   end
 
+  def service_resource
+    @service_resource ||= service_resources.find_by(:resource => self)
+  end
+
+  def delete_job
+    @delete_job ||= ManageIQ::Providers::EmbeddedTerraform::AutomationManager::Job.find_by(:target_id => id)
```
agrare: Minor but we should check the `:target_class => self.class.name` also

agrare: Also can we include anything else to differentiate a Retire Job from e.g. a Reconfigure Job in the future?

app/models/manageiq/providers/embedded_terraform/automation_manager/stack.rb
putmanoj: Oops ,  wrong name is comment,   should have been 'ServiceTerrraformTemplate'

 ServiceEmbeddedTerraform -> Workflow implementation - which uses `raw_delete_stack` method, 
 
 ServiceTerraformTemplate -> Automate implementation - whereas handles retire itself,  should not use `raw_delete_stack` method.

app/models/manageiq/providers/embedded_terraform/automation_manager/provision/state_machine.rb
putmanoj: ` terraform_runner_stack_job_id` might be useful for Reconfiguration,  
 - `terraform_runner_stack_job_id` is the `id` of the jobs run for the Stack, whereas `stack_id` indentifies the`Stack`.  

For now not so important

agrare: 👍 makes sense

app/models/manageiq/providers/embedded_terraform/automation_manager/stack.rb
putmanoj: Ok, though not sure which column can be used in Job, for the same.

app/models/manageiq/providers/embedded_terraform/automation_manager/stack/status.rb
```diff
@@ -1,11 +1,35 @@
...
+    self.stack = stack
+    miq_task   = stack.delete_miq_task.presence || stack.miq_task
 
-  def initialize(miq_task)
     super(miq_task.state, miq_task.message)
     self.task_status = miq_task.status
   end
 
+  def retiring?
+    stack.retiring? || (stack.delete_miq_task.present? && running?)
```
Fryguy, line 18: delete_miq_task reads like a verb/noun (a command) as opposed to adjective/noun - might be good to rename. (not for this PR)

app/models/manageiq/providers/embedded_terraform/automation_manager/job.rb
```diff
@@ -11,6 +11,13 @@ def self.create_job(template, env_vars, job_vars, credentials, action: ResourceA
     )
   end
 
+  def target=(target)
+    return if target.nil?
+
+    self.target_class = target.class.name
+    self.target_id = target.id
+  end
```
Fryguy, line 17: Why do we have this? If this is a polymorphic, then this method should exist on the base class anyway.

app/models/manageiq/providers/embedded_terraform/automation_manager/stack.rb
```diff
@@ -25,14 +27,16 @@ def create_stack(terraform_template, options = {})
         :miq_task                     => miq_task,
         :status                       => miq_task&.state,
         :start_time                   => miq_task&.started_on
-      )
+      ).tap do |stack|
+        job.target = stack
+        job.save!
```
Fryguy: Very minor

```suggestion
        job.update!(:target => stack)
```

app/models/manageiq/providers/embedded_terraform/automation_manager/stack.rb
```diff
@@ -48,19 +52,75 @@ def collect_authentications(manager, options)
...
-      self.finish_time = raw_status.completed? ? miq_task.updated_on : nil
-      save!
+    # when retired or retirement-failed, no further action required
+    return if retired? || error_retiring?
+
+    # where retirement is running
+    if retiring?
+      return if delete_miq_task.nil?
+
+      if raw_status.running? # delete_job&.is_active?
```
Fryguy, line 95: What does this comment mean?

app/models/manageiq/providers/embedded_terraform/automation_manager/stack.rb
```diff
@@ -105,16 +165,53 @@ def raw_stdout_html
...
+    if terraform_stack_id.blank?
+      $embedded_terraform_log.warn("Unable to retrieve stack data for stack(#{id}): terraform_stack_id is blank in job options")
+      return
+    end
 
     Terraform::Runner.stack(terraform_stack_id)
   end
 end
+
+def handle_stack_operation_error(operation, err)
```
Fryguy: Is this a global method outside of the module?

spec/models/manageiq/providers/embedded_terraform/automation_manager/stack_spec.rb
```diff
@@ -172,5 +172,126 @@
         end
       end
     end
+    describe "#raw_delete_stack" do
```
Fryguy, line 176:
```suggestion

    describe "#raw_delete_stack" do
```

app/models/manageiq/providers/embedded_terraform/automation_manager/job.rb
agrare: `::Job` doesn't have a `polymorphic` target, there is a `target_entity` reader [[ref]](https://github.com/ManageIQ/manageiq/blob/master/app/models/job.rb#L144-L146) but no writer

spec/models/manageiq/providers/embedded_terraform/automation_manager/provision_spec.rb
```diff
@@ -128,4 +130,66 @@
...
+
+    before do
+      subject.phase_context[:stack_id] = new_stack.id
+      new_stack.update(:miq_task => miq_task)
+      miq_task.update(:job => job)
+      allow(new_stack).to receive(:raw_status).and_return(new_stack.class.status_class.new(new_stack))
+    end
+
+    context "when all data is present" do
+      before { service_resource }
```
agrare: You can use `let!(:service_resource)` above so you don't have to initialize this in a before block

app/models/manageiq/providers/embedded_terraform/automation_manager/job.rb
Fryguy: Ok, then I recommend we move this method into the base Job model, but I'm ok with it as is.

Fryguy: Also target = nil no actually setting the underlying id to nil (and the class maybe) is a little weird, but fine for now.

agrare: Fixed target=nil should nil out class and id

app/models/manageiq/providers/embedded_terraform/automation_manager/stack.rb
putmanoj: missed cleaning that up ...
it was another condition, which also could have used,

agrare: @putmanoj can you put in a PR to delete it then?

PR #126, opened 2026-06-03
[WIP] Handle Terraform Runner Unavailability during migration or restarts(OOM)

putmanoj:
### Problem

The Terraform Runner service can become temporarily unavailable during:
- **Data migration after upgrade** — cipher migration at boot can take 7+ minutes with 10,000 records, returning HTTP 503 while running
- **OOM restarts** — the pod/container is killed and restarted, causing connection failures
- **Network transients** — brief connectivity issues between ManageIQ and the runner service

**Before this PR**, when the Terraform Runner was unreachable, the ManageIQ worker that made the API call would be fail — there was no availability check, no retry strategy, and the worker thread was stuck waiting on the HTTP timeout. 

### Solution

Introduce a **non-blocking, queue-based resilience strategy** that keeps workers free at all times:

1. **Fail-fast HTTP layer** — `Terraform::Runner` now makes exactly one HTTP attempt per API call. On 503 or connection/timeout failure, it raises `Terraform::Runner::TemporarilyUnavailable` immediately — no sleep, no retry loop.

2. **Availability check with caching** — A new `Terraform::Runner.available?` method checks the `/ready` health endpoint. Results are cached:
   - `true` is cached indefinitely (until a failed API call resets it)
   - `false` is cached for a configurable TTL (default 30s) to avoid hammering `/ready`

3. **Queue-based requeue at the job level** — Before every Runner API call, jobs/tasks check `available?`. If unavailable, they requeue themselves via MiqQueue with a 30-second delay and return immediately, freeing the worker.

4. **TTL-based give-up** — Each requeue path tracks a `runner_wait_started_at` timestamp. If the total elapsed wait exceeds a configurable max (default 10 minutes), the job aborts with a clear error rather than retrying forever.

5. **Race-condition protection** — Even if `available?` returns `true`, the actual API call can still fail if the service goes down between the check and the POST. All API-calling methods rescue `TemporarilyUnavailable` and requeue.

### Changes

**`lib/terraform/runner.rb`**
- Add `Terraform::Runner::TemporarilyUnavailable` exception class
- Rewrite `available?` with mutex-protected availability check + configurable cache TTL for `false` results
- Add `reset_available!` public method for cache invalidation
- Replace direct `terraform_runner_client.post(...)` calls with a new private `post` method that raises `TemporarilyUnavailable` on 503 / `Faraday::ConnectionFailed` / `Faraday::TimeoutError`
- Add `availability_max_wait_time` (public) — configurable max wait for requeue callers
- Add `availability_cache_ttl` (private) — configurable cache duration for `false`
- Thread-safe with `Mutex` synchronization

**`app/models/.../automation_manager/job.rb`**
- Add `poll_execute` state — inserted between `pre_execute` and `execute`; checks runner availability before proceeding
- `execute` — pre-checks `available?`, rescues `TemporarilyUnavailable`
- `poll_runner` — pre-checks `available?`, rescues `TemporarilyUnavailable` (from `retrieve_stack`)
- Add `requeue_or_abort_on_runner_unavailable` helper — timestamps, checks TTL, requeues or aborts
- Add `:poll_execute` to state machine transitions
- Reduce default `poll_interval` from 1 minute to 30 seconds

**`app/models/.../automation_manager/provision/state_machine.rb`**
- `run_provision` — checks `available?` before signaling `:provision`; requeues phase if unavailable
- `provision` — pre-checks `available?`, rescues `TemporarilyUnavailable` from `Stack.create_stack`
- Add `requeue_or_abort_on_runner_unavailable` — uses `phase_context[:runner_wait_started_at]`, TTL-based abort
- Fix error handling: replace `raise MiqException::MiqProvisionError` with `update_and_notify_parent` + `signal :finish` (proper pattern for Provision tasks)

**`app/models/.../automation_manager/stack.rb`**
- Add `delete_stack` method — wraps `raw_delete_stack` with availability check
- `raw_delete_stack` — rescues `TemporarilyUnavailable` and requeues
- Add `queue_delete_stack` — schedules retry via `MiqQueue` with configurable delay
- Add `requeue_or_fail_on_runner_unavailable` — timestamps, checks TTL, requeues or raises error

### Configuration

| Environment Variable | Default | Description |
|---|---|---|
| `TERRAFORM_RUNNER_AVAILABILITY_CACHE_TTL` | `30` (seconds) | How long to cache `available? == false` before re-checking `/ready` |
| `TERRAFORM_RUNNER_AVAILABILITY_MAX_WAIT_TIME` | `600` (seconds) | Maximum time a job will wait (via requeue cycles) before aborting |

### Flow

```
Job enters poll_execute / execute / poll_runner
  ├─ Terraform::Runner.available?
  │   ├─ true (cached or fresh /ready check)
  │   │   └─ Proceed with API call
  │   │       ├─ Success → continue state machine
  │   │       └─ 503 / ConnectionFailed / TimeoutError
  │   │           └─ Raise TemporarilyUnavailable
  │   │               └─ Reset availability cache
  │   │               └─ requeue_or_abort (check TTL)
  │   └─ false (cached within TTL — no HTTP call made)
  │       └─ requeue_or_abort (check TTL)
  │           ├─ Within max wait → requeue with 30s delay
  │           └─ Exceeded max wait → abort job with error
```

### Key Design Decisions

- **No sleep/blocking anywhere** — all waiting happens via MiqQueue delivery scheduling
- **Workers stay free** — a provisioning delay does not block UI requests
- **Availability-cache prevents thundering herd** — when runner is down, N concurrent jobs all return `false` from cache without N HTTP calls to `/ready`
- **Single retry point** — requeue logic is centralized in `requeue_or_abort_on_runner_unavailable` per component
- **Configurable timeouts** — operators can tune both cache TTL and max wait time for their environment

### Testing

All specs pass: **378 examples, 0 failures**

Specs added for:
- `available?` caching behavior (positive cache, negative cache with TTL, expiry)
- `TemporarilyUnavailable` raised on 503, ConnectionFailed, TimeoutError (for both `run` and `parse_template_variables`)
- `reset_available!` cache invalidation
- Job `poll_execute` / `execute` / `poll_runner` — requeue on unavailability, abort on TTL exceeded, rescue TemporarilyUnavailable
- Provision `run_provision` — requeue and TTL abort
- Stack stubs for `available?`

### DIFF
--- a/app/models/manageiq/providers/embedded_terraform/automation_manager/job.rb
+++ b/app/models/manageiq/providers/embedded_terraform/automation_manager/job.rb
@@ -27,14 +27,20 @@ class ManageIQ::Providers::EmbeddedTerraform::AutomationManager::Job < Job
 
   def poll_execute
     if Terraform::Runner.available?
+      options.delete(:runner_wait_started_at)
+      save!
       signal(:execute)
     else
-      $embedded_terraform_log.info("Terraform::Runner service is not available, requeueing poll_execute for Job#{id}")
-      queue_signal(:poll_execute, :deliver_on => Time.now.utc + poll_interval)
+      requeue_or_abort_on_runner_unavailable(:poll_execute)
     end
   end
 
   def execute
+    unless Terraform::Runner.available?
+      requeue_or_abort_on_runner_unavailable(:poll_execute)
+      return
+    end
+
     template_path = File.join(options[:git_checkout_tempdir], template_relative_path)
     credentials   = Authentication.where(:id => options[:credentials])
     action        = options[:action]
@@ -66,6 +72,9 @@ class ManageIQ::Providers::EmbeddedTerraform::AutomationManager::Job < Job
     save!
 
     queue_poll_runner
+  rescue Terraform::Runner::TemporarilyUnavailable => e
+    $embedded_terraform_log.warn("Terraform::Runner became unavailable during execute for Job#{id}: #{e.message}")
+    requeue_or_abort_on_runner_unavailable(:poll_execute)
   end
 
   def poll_runner
@@ -74,12 +83,16 @@ class ManageIQ::Providers::EmbeddedTerraform::AutomationManager::Job < Job
         $embedded_terraform_log.debug("Stack is still running, requeueing polling for Job#{id}")
         queue_poll_runner
       else
+        options.delete(:runner_wait_started_at)
+        save!
         signal(:post_execute)
       end
     else
-      $embedded_terraform_log.info("Terraform::Runner service is not available, requeueing polling for Job#{id}")
-      queue_poll_runner
+      requeue_or_abort_on_runner_unavailable(:poll_runner)
     end
+  rescue Terraform::Runner::TemporarilyUnavailable => e
+    $embedded_terraform_log.warn("Terraform::Runner became unavailable during poll_runner for Job#{id}: #{e.message}")
+    requeue_or_abort_on_runner_unavailable(:poll_runner)
   end
 
   def post_execute
@@ -134,6 +147,10 @@ class ManageIQ::Providers::EmbeddedTerraform::AutomationManager::Job < Job
     options.fetch(:poll_interval, 30.seconds).to_i
   end
 
+  def max_runner_wait_time
+    Terraform::Runner.availability_max_wait_time
+  end
+
   private
 
   def template
@@ -165,6 +182,20 @@ class ManageIQ::Providers::EmbeddedTerraform::AutomationManager::Job < Job
 
   def queue_poll_runner
     queue_signal(:poll_runner, :deliver_on => Time.now.utc + poll_interval)
+  end
+
+  def requeue_or_abort_on_runner_unavailable(signal_name)
+    options[:runner_wait_started_at] ||= Time.now.utc
+    save!
+
+    elapsed = Time.now.utc - options[:runner_wait_started_at]
+    if elapsed >= max_runner_wait_time
+      $embedded_terraform_log.error("Terraform::Runner unavailable for #{elapsed.to_i}s (max #{max_runner_wait_time}s), aborting Job#{id}")
+      abort_job("Terraform runner unavailable for too long", "error")
+    else
+      $embedded_terraform_log.info("Terraform::Runner not available, requeueing #{signal_name} for Job#{id} (waited #{elapsed.to_i}s/#{max_runner_wait_time}s)")
+      queue_signal(signal_name, :deliver_on => Time.now.utc + poll_interval)
+    end
   end
 
   def checkout_git_repository
--- a/app/models/manageiq/providers/embedded_terraform/automation_manager/job.rb
+++ b/app/models/manageiq/providers/embedded_terraform/automation_manager/job.rb
@@ -22,13 +22,15 @@ class ManageIQ::Providers::EmbeddedTerraform::AutomationManager::Job < Job
 
   def pre_execute
     checkout_git_repository
+    queue_signal(:poll_execute)
+  end
 
+  def poll_execute
     if Terraform::Runner.available?
       signal(:execute)
     else
-      delay = 10.seconds # poll_interval
-      $embedded_terraform_log.debug("Terraform::Runner service is not available, requeueing job pre_execute with delay: #{delay}")
-      queue_signal(:pre_execute, :deliver_on => Time.now.utc + delay)
+      $embedded_terraform_log.info("Terraform::Runner service is not available, requeueing poll_execute")
+      queue_signal(:poll_execute, :deliver_on => Time.now.utc + poll_interval)
     end
   end
 
@@ -75,7 +77,7 @@ class ManageIQ::Providers::EmbeddedTerraform::AutomationManager::Job < Job
         signal(:post_execute)
       end
     else
-      $embedded_terraform_log.debug("Terraform::Runner service is not available, requeueing polling")
+      $embedded_terraform_log.info("Terraform::Runner service is not available, requeueing polling")
       queue_poll_runner
     end
   end
@@ -117,6 +119,7 @@ class ManageIQ::Providers::EmbeddedTerraform::AutomationManager::Job < Job
       :initializing => {'initialize'       => 'waiting_to_start'},
       :start        => {'waiting_to_start' => 'pre_execute'},
       :pre_execute  => {'pre_execute'      => 'execute'},
+      :poll_execute => {'execute'          => 'execute'},
       :execute      => {'execute'          => 'running'},
       :poll_runner  => {'running'          => 'running'},
       :post_execute => {'running'          => 'post_execute'},
@@ -128,7 +131,7 @@ class ManageIQ::Providers::EmbeddedTerraform::AutomationManager::Job < Job
   end
 
   def poll_interval
-    options.fetch(:poll_interval, 1.minute).to_i
+    options.fetch(:poll_interval, 30.seconds).to_i
   end
 
   private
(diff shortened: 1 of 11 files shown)


lib/terraform/runner.rb
```diff
@@ -156,6 +165,8 @@ def stack_tenant_id
       end
 
       def run_terraform_runner_stack_api(request)
+        wait_for_runner_availability
```
agrare: If this is going to raise an exception it should be `wait_for_runner_availability!`

agrare (changes requested):
This will lock up a generic worker for 10 minutes for each time this is called while the worker is down.

Prefer using a job dispatcher and/or requeue in the job state machine so we're not just looping

app/models/manageiq/providers/embedded_terraform/automation_manager/job.rb
```diff
@@ -22,7 +22,16 @@ def start
...
-    signal(:execute)
+    queue_signal(:poll_execute)
+  end
+
+  def poll_execute
+    if Terraform::Runner.available?
+      signal(:execute)
+    else
+      $embedded_terraform_log.info("Terraform::Runner service is not available, requeueing poll_execute")
+      queue_signal(:poll_execute, :deliver_on => Time.now.utc + poll_interval)
```
agrare: Can you use `requeue_phase` here ?

app/models/manageiq/providers/embedded_terraform/automation_manager/job.rb
```diff
@@ -115,7 +131,8 @@ def load_transitions
   end
 
   def poll_interval
-    options.fetch(:poll_interval, 1.minute).to_i
+    # options.fetch(:poll_interval, 1.minute).to_i
```
agrare:
```suggestion
```

app/models/manageiq/providers/embedded_terraform/automation_manager/job.rb
agrare: Sorry `requeue_phase` is only in `MiqRequestTask::StateMachine` not in `Job`.

app/models/manageiq/providers/embedded_terraform/automation_manager/stack.rb
```diff
@@ -57,6 +57,15 @@ def retireable?
...
 
+  def delete_stack
+    if Terraform::Runner.available?
+      raw_delete_stack
+    else
+      $embedded_terraform_log.info("Terraform::Runner service is not available, requeue delete_stack")
+      requeue_phase
+    end
+  end
+
```
agrare, line 68: Is this covered by specs?
I don't think `requeue_phase` is valid here

app/models/service_embedded_terraform_mixin.rb
```diff
@@ -177,7 +177,9 @@ def credential_id_from_workflow_provision_request(request_options)
     credential_id = request_options[:credential_id]
 
     # If for provision_workflow, coming from customize tab, it will have a array like [id, name], and we only take id
-    credential_id = credential_id.first if credential_id.kind_of?(Array) && credential_id.first.present?
+    if credential_id.kind_of?(Array)
+      credential_id = credential_id.first.presence
+    end
```
agrare: This change looks unrelated to the rest of this PR

putmanoj: Yes, when working on current issue,  was seeing this problem with credential, so had to fix it.

Now moved to separate PR https://github.com/ManageIQ/manageiq-providers-embedded_terraform/pull/131

app/models/manageiq/providers/embedded_terraform/automation_manager/stack.rb
putmanoj: Yes it is covered with spec.

:-)  And it works perfectly with specs here https://github.com/putmanoj/manageiq-providers-embedded_terraform/blob/before-api-call-check-terraform-runner-availability/spec/models/manageiq/providers/embedded_terraform/automation_manager/stack_spec.rb#L176

Though, the re-queuing here might not be needed at all,  verifying again ...

agrare: Have you tested it live?  Your spec is stubbing the method, if I comment out the `#expect(stack).to receive(:requeue_phase)` line then it fails with
```
Failures:

  1) ManageIQ::Providers::EmbeddedTerraform::AutomationManager::Stack#raw_stdout_via_worker #delete_stack when Terraform::Runner is not available logs a message and requeues the phase
     Failure/Error: requeue_phase
     
     NameError:
       undefined local variable or method `requeue_phase' for #<ManageIQ::Providers::EmbeddedTerraform::AutomationManager::Stack:0x00007fe98d0c9e40>
```

putmanoj: In my live testing , where retiring,  It never reached the `requeue_phase `,  ..  as  usually availability status in cache was `true`.

So when I put terraform-runner service in migration state,  `Terraform::Runner` saw 503 error,  and went into `wait_for_runner_availability!`   where it can wait upto 600 seconds,   and when migration completed before the max wait time, everywhere worked good.

Anyways, I again force pushed, removed the particular commit,  verifying again,  might push a new commit

.github/workflows/ci.yaml
```diff
@@ -34,7 +34,7 @@ jobs:
       PGHOST: localhost
       PGPASSWORD: smartvm
     steps:
-    - uses: actions/checkout@v6
+    - uses: actions/checkout@v7
```
agrare, line 37: I don't think this was intended to be added to this branch

app/models/service_embedded_terraform_mixin.rb
```diff
@@ -27,7 +27,7 @@ def stack_opts(action = ResourceAction::PROVISION, overrides = {})
     stack_opts = {:action => action}.merge(input_vars_from_dialog(options.merge(overrides)))
 
     if instance_of?(ServiceEmbeddedTerraform)
-      stack_opts[:credential_id] = credential_id_from_workflow_provision_request(overrides)
+      stack_opts[:credential_id] = credential_id(overrides)
```
agrare, line 30: This needs to be rebased on master I'm guessing since we already merged this

lib/terraform/runner.rb
```diff
@@ -7,12 +7,17 @@ module Terraform
   class Runner
     class << self
       def available?
-        return @available if defined?(@available)
+        return @available if defined?(@available) && @available == true
```
agrare: Is there a reason why `return true if @available` wouldn't work?  And then `@available = nil` when you want to "reset" it?

lib/terraform/runner.rb
```diff
@@ -176,6 +184,74 @@ def run_terraform_runner_stack_api(request)
         end
       end
 
+      # Post to terraform-runner API with retry logic for 503 errors and connection failures
+      #
+      # @param action_endpoint [String] The API endpoint to post to
+      # @param request [Terraform::Runner::Request] The request object containing the payload
+      # @return [Faraday::Response] The HTTP response from the API
+      def post_with_retry(action_endpoint, request)
+        wait_for_runner_availability!
```
agrare: I'm confused, won't this run us into the same problem as this PR had originally where this will lock up a worker waiting for this?

.github/workflows/ci.yaml
putmanoj: Yes, not intended, removed now

app/models/service_embedded_terraform_mixin.rb
putmanoj: done now

lib/terraform/runner.rb
putmanoj: @agrare 
I have added a doc explaining the scenarios, this PR is trying to address -> https://github.com/putmanoj/manageiq-providers-embedded_terraform/blob/before-api-call-check-terraform-runner-availability/terraform-runner-availability-handling.md.  
I hope it makes things more clearer.

Also, let me setup a meeting later today,  if possible today,  lets discuss to bring closure on this PR.

Fryguy:
Merged https://github.com/ManageIQ/manageiq-providers-embedded_terraform/pull/134

PR #128, opened 2026-06-08
Fix retirement process for Embedded Terraform stacks with proper key transformation

putmanoj:
# Summary

This pull request fixes the retirement process for Embedded Terraform stacks by ensuring that job options are properly transformed to use symbol keys and improves memoization patterns for better handling of nil values.

### 🔧 Core Changes
- **Key Transformation**: Updated `raw_delete_stack` to transform hash keys from strings to symbols using `.transform_keys(&:to_sym)` when slicing `input_vars` and `credentials` from service_resource options
- **Improved Memoization**: Enhanced `service_resource` and `delete_job` methods to properly handle nil values by checking if instance variables are defined

### ✅ Test Coverage
- Updated existing stack specs to expect symbol keys in retirement options
- Added comprehensive retirement data preparation specs in provision_spec.rb
- Verified that provisioning properly stores data needed for retirement

## Files Changed

📄 **app/models/manageiq/providers/embedded_terraform/automation_manager/stack.rb**
- Line 70: Added `.transform_keys(&:to_sym)` to ensure consistent symbol keys
- Lines 168-170, 173-175: Improved memoization pattern using `defined?(@variable)`

📄 **spec/models/manageiq/providers/embedded_terraform/automation_manager/stack_spec.rb**
- Line 262-264: Updated expected options to use symbol keys (`:input_vars`, `:credentials`)

📄 **spec/models/manageiq/providers/embedded_terraform/automation_manager/provision_spec.rb**
- Lines 156-177: Added test for storing input_vars and credentials during provisioning
- Lines 191-245: Added comprehensive retirement data preparation specs

### DIFF
--- a/spec/models/manageiq/providers/embedded_terraform/automation_manager/provision_spec.rb
+++ b/spec/models/manageiq/providers/embedded_terraform/automation_manager/provision_spec.rb
@@ -136,6 +136,7 @@ describe ManageIQ::Providers::EmbeddedTerraform::AutomationManager::Provision do
     let(:job) { FactoryBot.create(:embedded_terraform_job, :options => job_options) }
     let(:job_options) { {:terraform_stack_id => "c247b890-4af1-11f1-bd0c-0f4596b0f2c6", :terraform_stack_job_id => "1"} }
     let!(:service_resource) { FactoryBot.create(:service_resource, :service => service, :resource => new_stack) }
+    let!(:auth) { FactoryBot.create(:authentication) }
 
     before do
       subject.phase_context[:stack_id] = new_stack.id
@@ -158,13 +159,16 @@ describe ManageIQ::Providers::EmbeddedTerraform::AutomationManager::Provision do
           "dialog_var1" => "value1",
           "dialog_var2" => "value2"
         }
-        subject.options[:credential_id] = 123
+        subject.options[:credential_id] = auth.id
+
+        # Mock the authentication for credential translation
+        allow(Authentication).to receive(:find).with(auth.id).and_return(auth)
 
         # Update service_resource with provision options
         service_resource.update!(
           :options => {
             "input_vars"  => {"var1" => "value1", "var2" => "value2"},
-            "credentials" => [123]
+            "credentials" => [auth.id]
           }
         )
 
@@ -172,7 +176,7 @@ describe ManageIQ::Providers::EmbeddedTerraform::AutomationManager::Provision do
 
         service_resource.reload
         expect(service_resource.options["input_vars"]).to eq({"var1" => "value1", "var2" => "value2"})
-        expect(service_resource.options["credentials"]).to eq([123])
+        expect(service_resource.options["credentials"]).to eq([auth.id])
         expect(service_resource.options["terraform_runner_stack_id"]).to eq(job_options[:terraform_stack_id])
       end
     end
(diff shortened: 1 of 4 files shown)


spec/models/manageiq/providers/embedded_terraform/automation_manager/provision_spec.rb
```diff
@@ -152,6 +152,34 @@
...
+        # Set up provision options with input_vars and credentials
+        subject.options[:dialog] = {
+          "dialog_var1" => "value1",
+          "dialog_var2" => "value2"
+        }
+        subject.options[:credential_id] = 123
+
+        # Mock the authentication for credential translation
+        auth = double("Authentication", :native_ref => "cred-ref-123")
+        allow(Authentication).to receive(:find).with(123).and_return(auth)
```
agrare: Wouldn't it be easier to just `let!(:auth) {}` and use `auth.id` and not have to mock `.find`
Then we don't have a made up ID and the tests should be more resilient to internal change

spec/models/manageiq/providers/embedded_terraform/automation_manager/provision_spec.rb
```diff
@@ -187,4 +215,63 @@
...
+    end
+    let!(:service_resource) do
+      FactoryBot.create(
+        :service_resource,
+        :service  => service,
+        :resource => new_stack,
+        :name     => "Provision",
+        :options  => {
+          "input_vars"  => {"region" => "us-east-1", "instance_type" => "t2.micro"},
+          "credentials" => [789]
```
agrare: Same thing here we could `let!(:auth)` and use a real ID

spec/models/manageiq/providers/embedded_terraform/automation_manager/provision_spec.rb
```diff
@@ -187,4 +215,63 @@
...
+        }
+      )
+    end
+
+    before do
+      subject.phase_context[:stack_id] = new_stack.id
+      new_stack.update(:miq_task => miq_task)
+      miq_task.update(:job => job)
+      allow(new_stack).to receive(:raw_status).and_return(new_stack.class.status_class.new(new_stack))
```
agrare: Why do we need to mock this?  Shouldn't it work naturally

spec/models/manageiq/providers/embedded_terraform/automation_manager/provision_spec.rb
putmanoj: Actually, we would need auth,  the stage for which, the test is,  it does not access Authentication, it only storing the value, ie  a id ( of Authentication ).  so removed it completely

spec/models/manageiq/providers/embedded_terraform/automation_manager/provision_spec.rb
putmanoj: Same as above,

spec/models/manageiq/providers/embedded_terraform/automation_manager/provision_spec.rb
```diff
@@ -152,6 +152,33 @@
...
+      it "stores input_vars and credentials for retirement" do
+        # Set up provision options with input_vars and credentials
+        subject.options[:dialog] = {
+          "dialog_var1" => "value1",
+          "dialog_var2" => "value2"
+        }
+        subject.options[:credential_id] = auth.id
+
+        # Mock the authentication for credential translation
+        allow(Authentication).to receive(:find).with(auth.id).and_return(auth)
```
agrare: Do you need this?

agrare: This isn't needed, I deleted it locally and tests still pass.

```suggestion
```

agrare (approved):
I have a commit in this branch but it is extremely minor.

PR #131, opened 2026-06-16
Fix credential_id for provision workflow

putmanoj:
When credentials coming from **Customize** tab, for provision_workflow, the credential data, will have a array like [id, name], and we only return id, else return nil.

Provisioning fails, when no credentials is selected or required for the catalog item,  this fix resolves the problem.

### DIFF
--- a/app/models/service_embedded_terraform_mixin.rb
+++ b/app/models/service_embedded_terraform_mixin.rb
@@ -27,7 +27,7 @@ module ServiceEmbeddedTerraformMixin
     stack_opts = {:action => action}.merge(input_vars_from_dialog(options.merge(overrides)))
 
     if instance_of?(ServiceEmbeddedTerraform)
-      stack_opts[:credential_id] = credential_id_from_workflow_provision_request(overrides)
+      stack_opts[:credential_id] = credential_id(overrides)
     end
 
     translate_credentials!(stack_opts)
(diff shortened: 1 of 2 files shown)


app/models/service_embedded_terraform_mixin.rb
```diff
@@ -177,7 +177,9 @@ def credential_id_from_workflow_provision_request(request_options)
     credential_id = request_options[:credential_id]
 
     # If for provision_workflow, coming from customize tab, it will have a array like [id, name], and we only take id
-    credential_id = credential_id.first if credential_id.kind_of?(Array) && credential_id.first.present?
+    if credential_id.kind_of?(Array)
+      credential_id = credential_id.first.presence
+    end
```
agrare: We have a helper method for exactly this, let me see if it will work where this is called from
`get_value(credential_id)` is in `MiqRequestWorkflow` and `get_option` is in `MiqRequestMixin` and `ServiceOrchestrationOptionsMixin` the latter might work.

agrare: Yeah its not defined in this class, I think we should match that behavior for consistency though maybe define a `get_value` here and then you can do `credential_id = get_value(request_options[:credential_id])`
```
  def get_value(data)
    data.kind_of?(Array) ? data.first : data 
  end
```

Also can you add a spec covering the case that you're fixing here

agrare (approved):

PR #134, opened 2026-08-05
Requeue state if `Terraform::Runner` is unavailable

agrare:
Alternative to https://github.com/ManageIQ/manageiq-providers-embedded_terraform/pull/126 which adds a retirement state machine so we can requeue in the state machine if the terraform runner is unavailable rather than simply sleep in a loop blocking a worker.

TODO:
- [x] Live test of Provision with `Terraform::Runner` initially unavailable
- [x] Live test of Retire with `Terraform::Runner` initially unavailable
- [x] Live test of Provision with `Terraform::Runner` initially available but becoming unavailable during state_machine
- [x] Live test of Retire with `Terraform::Runner` initially available but becoming unavailable during state_machine
- [x] Live test of Provision state machine timing out if `Terraform::Runner` is never available
- [x] Live test of Retire state machine timing out if `Terraform::Runner` is never available

Depends on:
- [x] https://github.com/ManageIQ/manageiq/pull/23927


app/models/manageiq/providers/embedded_terraform/automation_manager.rb
```diff
@@ -21,4 +21,8 @@ def self.catalog_types
       "embedded_terraform"         => N_("Terraform Template")
     }
   end
+
+  def self.orchestration_stack_retire_task_class
+    ManageIQ::Providers::EmbeddedTerraform::AutomationManager::OrchestrationStackRetireTask
```
Fryguy: Can (should?) this be

```suggestion
    self::OrchestrationStackRetireTask
```

lib/terraform/runner.rb
```diff
@@ -114,8 +125,16 @@ def parse_template_variables(template_path)
           *request.build_json_post_arguments
         )
 
+        if http_response.status == 503
+          available_clear_cache
+          raise TemporarilyUnavailable, "Terraform Runner service is temporarily unavailable (503)"
```
Fryguy, line 130: Does anything catch the TemporarilyUnavailable?

agrare: The `Retire` and `Provision` `StateMachine` checks it.  It wasn't implemented when you reviewed this but has since been added.  I'm trying to pull pieces out of https://github.com/ManageIQ/manageiq-providers-embedded_terraform/pull/126 in a way that makes sense and hadn't gotten to that yet.

app/models/manageiq/providers/embedded_terraform/automation_manager.rb
agrare: :+1: done

app/models/manageiq/providers/embedded_terraform/automation_manager/retire/state_machine.rb
```diff
@@ -0,0 +1,12 @@
+module ManageIQ::Providers::EmbeddedTerraform::AutomationManager::Retire::StateMachine
+  extend ActiveSupport::Concern
+
+  # If the Terraform::Runner isn't available then don't start the retirement process
```
Fryguy: Minor but this comment I think belongs on line 6 as opposed to the whole method.

app/models/manageiq/providers/embedded_terraform/automation_manager/terraform_runner_mixin.rb
```diff
@@ -0,0 +1,24 @@
...
+  end
+
+  def terraform_runner_unavailable_requeue_phase
+    # Track when the runner first became unavailable
+    phase_context[:terraform_runner_unavailable_since] ||= Time.now.utc
+
+    unavailable_duration = Time.now.utc - phase_context[:terraform_runner_unavailable_since]
+
+    if unavailable_duration > TERRAFORM_RUNNER_UNAVAILABLE_TIMEOUT
+      error_message = "Terraform Runner has been unavailable for #{unavailable_duration.to_i} seconds, exceeding timeout of #{TERRAFORM_RUNNER_UNAVAILABLE_TIMEOUT.to_i} seconds"
```
Fryguy: Should be translated? Not sure if this will be customer facing or not.

app/models/manageiq/providers/embedded_terraform/automation_manager/terraform_runner_mixin.rb
```diff
@@ -0,0 +1,24 @@
+module ManageIQ::Providers::EmbeddedTerraform::AutomationManager::TerraformRunnerMixin
+  TERRAFORM_RUNNER_UNAVAILABLE_TIMEOUT = 10.minutes
```
Fryguy: Should this be configurable? Not saying it should just raising the question - I can't think of a reason an end user would want to change it.

app/models/manageiq/providers/embedded_terraform/automation_manager/retire/state_machine.rb
agrare: Sure, that is the only reason the whole method is overridden which is why I had the comment on the method.

Fryguy: Ohhh I see.

agrare: I'll add a comment about marking the terraform runner as available as well, since this method does that as well.

Fryguy (approved):

